PR #40, opened 2017-09-20
Adding Dockerfile and corresponding documentation

ibnesayeed:
This Dockerfile will make development, testing, and usage of the tool easier while it would also allow automated build on Dockerhub on each change, both on commits and tags (such as on various versioned releases).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,54 @@
+ARG MAVEN_VERSION=3-jdk-8
+ARG JDK_VERSION=8-jre
+
+FROM maven:$MAVEN_VERSION AS builder
+
+ARG NOTEBOOK_VERSION=0.7.0-pre2
+ARG SCALA_VERSION=2.10.5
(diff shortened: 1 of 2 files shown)


ibnesayeed:
We certainly can make this PR against `docker-aut`, which will require some changes to pull the code of `aut` from this repo. However, I don't see a good reason to even keep a separate repo for docker image, unless there are some goals that I am not aware of.

Keeping this file here in the main repo would:

* help developers build and test the code utilizing docker as they make changes in this code
* enable auto-builds in DockerHub (or any other place that receives web hooks) on each change here in the main code
* does not add any complexity or artifacts in the code other than a self-contained Dockerfile

Publishing it elsewhere would:

* require manual changes in the other repo to trigger an auto-build each time this main repo is updated
* developer's docker-based workflow would be broken as even copying the other repo locally would not allow building image based on the modifications in the main code without modification in the Dockerfile

Issue #55, opened 2017-10-02
Create tests for WriteGDF.scala


ruebot:
[`WriteGDF.scala`](https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/spark/matchbox/WriteGDF.scala) has [no test coverage](https://codecov.io/gh/archivesunleashed/aut/src/master/src/main/scala/io/archivesunleashed/spark/matchbox/WriteGDF.scala). We need to create tests for it.

ruebot:
We have 90% coverage on this one. We should cover this conditional:
https://codecov.io/gh/archivesunleashed/aut/src/master/src/main/scala/io/archivesunleashed/spark/matchbox/WriteGDF.scala#L38

greebie:
Not sure how to cover the results here.  Either way, the function will return a Unit, (), so assertion will not work.  Nor can I test for the function not creating a file.  

Would it be better if the function returned an Option or a Bool?

greebie:
I think the solution for full coverage here is to ask the function to return a value on failure vs true.  This can be an option or a Bool.  (probably Bool).

ruebot:
Cool. If you want to toss in a PR, let's see how it goes.

greebie:
It'll probably blow up some other test, but we'll see.  :)

ruebot:
Breaking shit, and blowing stuff up is how I learn :smiley:

greebie:
🗡  💩 💥 💩 💩

ruebot:
Resolved with: https://github.com/archivesunleashed/aut/commit/64a48515f42cbdc7da2254341f30f3735bee03e6

Issue #60, opened 2017-10-02
Improve RecordLoader.scala test coverage


ruebot:
[`RecordLoader.scala`](https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/spark/matchbox/RecordLoader.scala) [test coverage](https://codecov.io/gh/archivesunleashed/aut/src/master/src/main/scala/io/archivesunleashed/spark/matchbox/RecordLoader.scala) can be improved.

greebie:
Going to start with this one for now.

ruebot:
Couple more lines to cover here that #84 did not cover:
https://codecov.io/gh/archivesunleashed/aut/src/master/src/main/scala/io/archivesunleashed/spark/matchbox/RecordLoader.scala

ruebot:
Partially addressed with: https://github.com/archivesunleashed/aut/commit/a539a1c96df275315c50f6c6f51a8a2aac4884f5

greebie:
(RemoveHTML is covered 100%)

greebie:
My bad here.  The RecordLoader test coverage is not 100%, it's just misnamed.  I will try to work on it today.

ruebot:
Partially addressed with: https://github.com/archivesunleashed/aut/commit/ffb23d8323eb32de85a5ba04c946d06283538ba3

ruebot:
`RecordLoader.scala` is gone, and integrated elsewhere. Closing this one, and will open up a new ticket.

Issue #59, opened 2017-10-02
Improve TupleFormatter.scala test coverage


ruebot:
[`TupleFormatter.scala`](https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/matchbox/TupleFormatter.scala) [test coverage](https://codecov.io/gh/archivesunleashed/aut/src/master/src/main/scala/io/archivesunleashed/matchbox/TupleFormatter.scala) can be improved.

ruebot:
Couple more lines to cover here that #84 did not cover:
https://codecov.io/gh/archivesunleashed/aut/src/master/src/main/scala/io/archivesunleashed/spark/matchbox/TupleFormatter.scala

ruebot:
Partially addressed with: https://github.com/archivesunleashed/aut/commit/aaad971cb6ab09c9280504afe4dcaa2364ff1921

greebie:
Tried a number of times to improve this coverage with no success.  Not sure how to test for the LowPriorityFlatten class.

Issue #57, opened 2017-10-02
Improve RemoveHttpHeader.scala test coverage


ruebot:
[`RemoveHttpHeader.scala`](https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/spark/matchbox/RemoveHttpHeader.scala) [test coverage](https://codecov.io/gh/archivesunleashed/aut/src/master/src/main/scala/io/archivesunleashed/spark/matchbox/RemoveHttpHeader.scala) can be improved.

ruebot:
Still a few more lines to cover: https://codecov.io/gh/archivesunleashed/aut/src/master/src/main/scala/io/archivesunleashed/spark/matchbox/RemoveHttpHeader.scala#L26

greebie:
This is 100% now.  Closing.

PR #84, opened 2017-10-04
First round of unit tests

greebie:
I figure it is worth getting people to look at these early, so I can get feedback on my general approach.

These are fairly simple tests that should cover most use cases in the matchbox.  Most are basic simple assertions.  The WriteGDF is a little bit more complicated because it requires instantiating an RDD and creates a file which must be deleted.

### DIFF
--- a/src/test/scala/io/archivesunleashed/spark/matchbox/RemoveHTMLTest.scala
+++ b/src/test/scala/io/archivesunleashed/spark/matchbox/RemoveHTMLTest.scala
@@ -1,3 +1,20 @@
+/*
+ * Archives Unleashed Toolkit (AUT):
+ * An open-source platform for analyzing web archives.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


src/test/scala/io/archivesunleashed/spark/matchbox/RemoveHTMLTest.scala
```diff
@@ -0,0 +1,22 @@
+package io.archivesunleashed.spark.matchbox
```
ruebot: Can you put the LICENSE header in here?

src/test/scala/io/archivesunleashed/spark/matchbox/RemoveHttpHeaderTest.scala
```diff
@@ -0,0 +1,14 @@
+package io.archivesunleashed.spark.matchbox
```
ruebot: Same here with the LICENSE.

ianmilligan1:
Great stuff @greebie!

greebie:
All added!  Hurray!

greebie:
Resolves #55, #56, #57, #59 & #60.

PR #94, opened 2017-10-15
Setup cobertura better so we have local html reports.

ruebot:
**GitHub issue(s)**:

* n/a/

Adds a cobertura reporting section, and removes unneeded config from build section for cobertura.

# How should this be tested?

Build locally with `mvn cobertura:cobertura` and once the build completes, you should have a `target/site/cobertura` directory. Open `index.html` and should could see something along the lines of this:

screenshot from 2017-10-15 14-47-57

Tag @greebie you might be very interested in this. Maybe @ianmilligan1 too?


ianmilligan1:
This looks great - very useful. And I think @greebie will love this!

Issue #72, opened 2017-10-02
Create tests for SocialMediaLinks.scala


ruebot:
[`SocialMediaLinks.scala`](https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/spark/scripts/SocialMediaLinks.scala) has [no test coverage](https://codecov.io/gh/archivesunleashed/aut/src/master/src/main/scala/io/archivesunleashed/spark/scripts/SocialMediaLinks.scala). We need to create tests for it.

ianmilligan1:
heh, good catch. Makes sense to me.

greebie:
Do you prefer moving these for better test coverage or leave them alone?

ianmilligan1:
Moved conversation over to AUT slack to see if we want to keep social media analysis in here (if any others want to chime in, feel free to do so on this issue or [join us over at the Slack](http://archivesunleashed.org).

ruebot:
Resolved with: https://github.com/archivesunleashed/aut/commit/c6fc97428bebfc1a2406340179175245d6a232e0

Issue #70, opened 2017-10-02
Create tests for CrawlStatistics.scala


ruebot:
[`CrawlStatistics.scala`](https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/spark/scripts/CrawlStatistics.scala) has [no test coverage](https://codecov.io/gh/archivesunleashed/aut/src/master/src/main/scala/io/archivesunleashed/spark/scripts/CrawlStatistics.scala). We need to create tests for it.

greebie:
See #72.

ruebot:
Resolved with: https://github.com/archivesunleashed/aut/commit/c6fc97428bebfc1a2406340179175245d6a232e0

PR #96, opened 2017-10-15
Remove example scripts. Resolves #95, #70, #71, #72.

### LINKED ISSUE #95: Are src/main/scala/io/archivesunleashed/spark/scripts needed?
@ianmilligan1 I'm looking at the test coverage reports, I noticed that there are some hard coded paths in the scripts here `master/src/main/scala/io/archivesunleashed/spark/scripts`.

Are these three files required? If so, they need to be generalized and put in `matchbox`, or they can just but moved over to `aut-docs` as examples. If they are not necessary, may I remove them?
* https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/spark/scripts/CrawlStatistics.scala
* https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/spark/scripts/Filter.scala
* https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/spark/scripts/SocialMediaLinks.scala

I see what @greebie was getting at now [here](https://github.com/archivesunleashed/aut/issues/72#issuecomment-335491771), and I believe the conversation in Slack was focused on Twitter support. So, we should make a final decision on these files here.
 
If the files can be removed, then we can also close:
* #70, #71, #72

ruebot:
**GitHub issue(s)**:

* #70 
* #71 
* #72 
* #95 

Remove unnecessary files.

# How should this be tested?

TravisCI turns green, we're good. CodeCov should get a nice bump too.

# Interested parties

@ianmilligan1 @greebie @lintool


ianmilligan1:
It’d be good for @lintool to merge this one if he agrees - he wrote these components if I remember correctly!

Issue #13, opened 2017-07-06
DataFrames support


lintool:
We need DataFrames support. Not sure if the best course of action is to resolve #12 and build PySpark DataFrames directly, or to build DataFrame support and hash out in Scala before making sure everything works in PySpark.

ianmilligan1:
I'm the least experienced here, but after seeing the various implementations of DataFrames in [Scala](https://docs.databricks.com/spark/latest/dataframes-datasets/introduction-to-dataframes-scala.html) and [Python](https://docs.databricks.com/spark/latest/dataframes-datasets/introduction-to-dataframes-python.html), it probably is more economical to resolve #12 and build PySpark DataFrames directly?

This is going to be a great feature, btw.

greebie:
What is the use case for DataFrames support via spark?  Filtering & sorting by categorical var?

The dataframe documentation in scala seems pretty straight forward.  I guess it just depends on how much a priority it is for the base.   

Another consideration is that if you want people to migrate to aut from warcbase, dataframe support could be nice selling feature.

lintool:
> What is the use case for DataFrames support via spark? Filtering & sorting by categorical var?

Correct. In our proposed FAAV cycle [1], DataFrames would support simplified easy filtering.

[1] http://dl.acm.org/citation.cfm?id=3097570

greebie:
So the status quo (no dataframes) would require users to map & filter into a derivative, output as delimited, Json or some flavor of XML and then pump that into pandas or R.  I take it the main problem with this is that Pandas could end up with useless garbage -- or worse, find that necessary info was filtered out. 

I assume supporting dataframes in Warcbase would capture problems at an earlier stage and then output to a nice data table that could be imported to pandas, R or some other useful format (plain old json would work for me).

I think having an example output using the dataframe would be nice.  I am ambivalent as to when, although if I really wanted to see it in action, I could try to just implement it in spark myself.  I think waiting for pyspark is probably the best option.

lintool:
PySpark and DataFrames are two orthogonal issues.

PySpark makes it easy to integrate AUT with other Python based tools. DataFrames makes the filtering process more intuitive. They make a powerful combo!

With DataFrames, we can do things like:

```
pages.filter(pages['url'].startswith('foo.com') and 'bar' in pages['contents']).count().show()
```

Translation: count all pages that are in the `foo.com` domain that contain the keyword `bar`.

More details: https://spark.apache.org/docs/latest/sql-programming-guide.html

[...]

Issue #12, opened 2017-07-06
PySpark support


lintool:
We need PySpark support.

ianmilligan1:
Assigned to @MapleOx, aiming to finish by end of October. Any questions in Slack or here. 😄

ianmilligan1:
OK we're in final review on #100 now, when that is merged this can be closed.

ianmilligan1:
@dhop is taking this issue on!

PR #100, opened 2017-10-20
PySpark support for core AUT functionality. #12, #13. 

MapleOx:
**GitHub issue(s)**:

* #12 
* #13 

Adds PySpark support for the core functionality of the Archives Unleashed Toolkit. Added:

* Ability to load arcs and warcs as RDDs or DataFrames in PySpark,
* Python DataFrame transformation functions like `keepValidPages` and `keepImages` that mimic the RDD transformations in AUT Scala,
* Python versions of several of the AUT matchbox functions, like `ExtractLinks` and `RemoveHTML`,
* Two example python scripts to demonstrate usage.

# How should this be tested?

First, run `mvn clean package` to rebuild AUT. After that, you can either run from the PySpark shell or submit your own script. 

**To run from the PySpark shell:**
1. `cd` into the `aut/` directory. 
2. Run `zip pyaut src/main/python/*.py` to make a zip of the Python files. 
3. Run the following command to start the PySpark shell:
```
pyspark --jars target/aut-0.10.1-SNAPSHOT-fatjar.jar --driver-class-path target/aut-0.10.1-SNAPSHOT-fatjar.jar --py-files python.zip
```
4. You can now use AUT in the PySpark shell!

**To run a script:**
1. `cd` into the `aut/` directory. 
2. Run `zip pyaut src/main/python/*.py` to make a zip of the Python files. 
3. Run the following command to submit your script:
```
spark-submit --jars target/aut-0.10.1-SNAPSHOT-fatjar.jar --driver-class-path target/aut-0.10.1-SNAPSHOT-fatjar.jar --py-files python.zip path/to/myscript.py 
```

Two example scripts can be found in the `src/main/python/scripts` directory. 

# Additional Notes:

The more complicated AUT Scala matchbox functions, like `ExtractGraph` and `ExtractEntities`, are not included in this PR. I plan to add them progressively. 

The Python code uses BeautifulSoup4, a library for extracting data from HTML and XML. 

# Interested parties

@lintool @ruebot @ianmilligan1 @greebie


ianmilligan1:
I'll leave reviews to @ruebot and @lintool, but I think it'd be good to get this into the repository so we (and others in the community) can explore PySpark using the main repo. It's sitting apart from the main Scala functions so wouldn't affect AUT's functionality.

src/main/python/DFTransformations.py
```diff
@@ -0,0 +1,130 @@
...
+from pyspark.sql.functions import udf
+from pyspark.sql.types import BooleanType
+import re
+
+def countItems(rdd): 
+  return rdd.map(lambda r: (r, 1)) \
+    .reduceByKey(lambda c1, c2: c1 + c2) \
+    .sortBy(lambda f: f[1], ascending = False)
+
+# def keepImages(df):
```
ruebot: Do we need this commented out code?

src/main/python/DFTransformations.py
```diff
@@ -0,0 +1,130 @@
...
+  def content_filter(content):
+    for pattern in contentREs:
+      if re.match(pattern, content) is not None:
+        return True
+    return False
+  content_filter_udf = udf(content_filter, BooleanType())
+  return df.filter(content_filter_udf(df['contentString']))
+
+
+# ---- TODO: All discard filtering operations ---- #
```
greebie: Is this still "TODO" or have you completed the task?  (remove comment)

src/main/python/DetectLanguage.py
```diff
@@ -0,0 +1,23 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from langdetect import detect
+
+def DetectLanguage(input: String):
+	if input == "":
+		return ""
```
greebie: I wonder if a try ... catch would work better here.  It seems like there are a few different reasons why a langdetect might burp besides empty string.

https://github.com/shuyo/language-detection/blob/master/src/com/cybozu/labs/langdetect/LangDetectException.java

src/main/scala/io/archivesunleashed/spark/pythonhelpers/RecordLoaderPythonHelper.scala
```diff
@@ -0,0 +1,45 @@
...
+  def loadWarc(path: String, jssc: JavaSparkContext, spark: SparkSession): DataFrame = {
+    val sc = jssc.sc
+    val rdd = RecordLoader.loadWarc(path, sc).keepValidPages()
+    val df = spark.createDataFrame(rdd = rdd, beanClass = classOf[ArchiveRecord])
+    df
+  }
+
+  def loadArchives(path: String, jssc: JavaSparkContext, spark: SparkSession): DataFrame = {
+    val sc = jssc.sc
+    val rdd = RecordLoader.loadArchives(path, sc).keepValidPages()
```
greebie: Bigger discussion, but is there a rationale for .keepValidPages() inside vs outside of the loadArchives function?  In Scala, we require people use it separately from the load function.

Does it make sense to support accessing invalid pages?  If not, maybe alter the scala script to match this?

ianmilligan1: Ping @lintool here

greebie: I think the approach here works okay for this PR, but we should add an Issue to address the difference in approaches as it might be confusing.

src/main/python/DFTransformations.py
greebie: I am willing and able to look at unit test coverage here once people have tried it and accept that it does what we want it to do.  From I've seen, pytest is the best library for handling a pyspark context:  https://stackoverflow.com/questions/33811882/how-do-i-unit-test-pyspark-programs

Will we need python dependency management here?

src/main/python/scripts/filterByDateScript.py
```diff
@@ -0,0 +1,37 @@
...
+from DFTransformations import *
+from ExtractDomain import ExtractDomain
+from ExtractLinks import ExtractLinks
+from ExtractDate import DateComponent
+from RemoveHTML import RemoveHTML
+from pyspark.sql import SparkSession
+
+if __name__ == "__main__":
+	# replace with your own path to archive file
+	path = "/Users/Prince/Projects/pyaut/aut/example.arc.gz"
```
greebie: Can't remember if there is an equivalent to the Resource Scala library for Python.  It would be better for unit testing / coverage.  If not, we could also move this script to aut-docs as we did with the Scala scripts.

src/main/python/DFTransformations.py
MapleOx: Good catch, that comment should be removed. I probably have some other TODOs comments that should be deleted too.

src/main/scala/io/archivesunleashed/spark/pythonhelpers/RecordLoaderPythonHelper.scala
MapleOx: Yeah, my rationale was that `.keepValidPages()` was always being called after loading the archive, so I might as well make `RecordLoader` do it automatically. That being said, I can remove this if it's undesirable and revert to how it's done in Scala.

src/main/python/scripts/filterByDateScript.py
MapleOx: Yeah, the scripts were just to give some examples of how to use AUT in PySpark, and it doesn't really matter where the scripts are located. I just put them under aut for convenience.

src/main/scala/io/archivesunleashed/spark/pythonhelpers/RecordLoaderPythonHelper.scala
greebie: Assuming @ruebot is okay with, I think it makes sense to move it to the RecordLoader as well.  I think that still means you need a invalid parameter check `(path, sc, invalid)` instead of including the .keepValidPages() call.

greebie: Also - I suggest letting me fix the RecordLoader because it's probably going to break a whole bunch of unit tests.  I'll get on it Sat, Sun or Mon.

ruebot: @greebie do you want to make a pull against @MapleOx's pull?

lintool: I'd rather not do that... let's get @MapleOx 's PR merged as soon as reasonably possible. We can start issues on what needs to be done next, and attach directly to this PR.

greebie: Yes.  @MapleOx did a great job, and this is more of a code refactor having nothing directly to do with the work.

I'll create an issue and branch off until the PR is merged.

src/main/python/DFTransformations.py
MapleOx: @ruebot By "Scala scripts", do you mean the things under `src/main/python/scripts/` like `extractLinkScript.py`, or the matchbox functions like `ExtractDomain`?

MapleOx:
@ruebot Okay, I'll remove the example scripts from this PR. 

For the matchbox functions that are used in RDD transformations, I rewrote them in Python because I was unable to call them straightforwardly from Scala. 

More details: I originally tried to just call them from Scala, like this:
```
def ExtractDomain(sc, url, source = ""):
  jvm = sc._jvm
  return jvm.io.archivesunleashed.spark.matchbox.ExtractDomain.apply(url, source)
```
However, this doesn't work if we call `ExtractDomain` in a transformation, like
```
rdd.map(lambda r => ExtractDomain(sc, r.url))
```
because of this:
```
pickle.PicklingError: Could not serialize object: Exception: It appears that you are attempting to 
reference SparkContext from a broadcast variable, action, or transformation. SparkContext can only
be used on the driver, not in code that it run on workers. For more information, see SPARK-5063.
```
I wasn't able to think of a way around this, so I just decided to write Python versions of the RDD transformation matchbox functions -- let me know if you guys have a better solution!

Now, for the matchbox functions that are not used in RDD transformations like `ExtractGraph`, it might be possible to just call them directly from Scala, but I haven't tried this yet. In any case, the Python matchbox functions we have right now are all ones that might be used inside RDD transformations.

greebie:
I think the issue is that RDD does not support nested spark transformations.  So you can create a dataframe and then run transformations, but you cannot run transformations and create a dataframe out of it.

Or something like that.  I don't understand it fully.  The example they give is trying to create two rdds and using one rdd to transform the other.  Leads to too many issues in large datasets I suppose.

ianmilligan1:
@greebie will test and then I think we're good to merge. 👍

ianmilligan1:
Better format to create zip, from `src/main/python` run

```
zip -r ~/pyspark/aut/pyspark1.zip .
```

Or some variety of that.

greebie:
Slightly better zipping script: 

` (cd src/main/python && zip -r - .) > pyaut.zip `
Saves the trouble and confusion of people having to find src/main/python/.

ianmilligan1:
FWIW I have been running it with a vanilla Jupyter notebook by running

```
PYSPARK_DRIVER_PYTHON=jupyter PYSPARK_DRIVER_PYTHON_OPTS=notebook ~/Dropbox/pyspark/spark-2.2.0-bin-hadoop2.7/bin/pyspark --jars target/aut-0.10.1-SNAPSHOT-fatjar.jar --driver-class-path target/aut-0.10.1-SNAPSHOT-fatjar.jar --py-files /Users/ianmilligan1/dropbox/pyspark/aut/pyspark1.zip
```

from the `aut` directory.

greebie:
Latest pull request fixes bugs I found in keepLanguages() while testing.

Issue #128, opened 2017-11-30
ERROR ArcRecordUtils - Read 1224 bytes but expected 1300 bytes


ruebot:
This is coming up a lot in the build.

Example:

```
2017-11-30 09:19:54,601 [Executor task launch worker for task 0] ERROR ArcRecordUtils - Read 1224 bytes but expected 1300 bytes. Continuing...
WARNING Record STARTING at 0 has 1761 trailing byte(s): file:/home/nruest/git/aut/target/test-classes/arc/badexample.arc.gz: {subject-uri=filedesc://IAH-20080430204825-00000-blackbook.arc, ip-address=0.0.0.0, origin=InternetArchive, length=1300, absolute-offset=0, creation-date=20080430204825, content-type=text/plain, version=1.1}
```

ianmilligan1:
We had this back in Warcbase - see issue here <https://github.com/lintool/warcbase/issues/199>.

According to this it should have been fixed in webarchive-commons (1.1.7). @jrwiebe did quite a bit of digging on this problem.

ruebot:
@anjackson any chance this resurfaced in webarchive-commons 1.1.8?

anjackson:
@ruebot Possibly, although [the changeset doesn't seem to affect the original fix](https://github.com/iipc/webarchive-commons/compare/webarchive-commons-1.1.7...webarchive-commons-1.1.8). Any chance you've got an older version on the classpath somewhere?

ruebot:
`+- org.netpreserve.commons:webarchive-commons:jar:1.1.8:compile` is the only `webarchive-commons` package that comes up with `mvn dependency:tree`.

anjackson:
Hmmm, I'm confused. Your version of [ArcRecordUtils](https://github.com/archivesunleashed/aut/blob/master/src/main/java/io/archivesunleashed/data/ArcRecordUtils.java#L72) doesn't have to additional code @jrwiebe [added to cope with this situation](https://github.com/lintool/warcbase/blob/7bdc8b55fdbf96a6fd5a246de761fb344563ce1f/src/main/java/org/warcbase/data/ArcRecordUtils.java#L61-L66).

Or am I misunderstanding the issue?

ruebot:
Oh, that weird.

ruebot:
[...]

PR #134, opened 2017-12-04
Suppresses error handling for ARCs w/ 76 character header difference

ianmilligan1:
This is a recurring error due to header problems. Temporary patches #128.

**GitHub issue(s)**:

* #128 

This is a minor pull request that basically suppresses errors when ingesting ARC files. Right now, due to issues discussed in #128, every ARC that we ingest says that it is finding 76 less characters than it is expecting.  

Until we have more cycles to substantially rework the ARC record handling, I think suppressing it is a good call.

# How should this be tested?

It should pass on Travis CI. It could also use another set of eyeballs. I have tested it on sample ARC files.

# Additional Notes:

The most important thing is that my knowledge of Java comes from doing the CodeAcademy tutorials this morning, so please keep that in mind. This is very much a rookie trying to take care of low-hanging fruit. But this error message has _bugged me like hell_ for the last two years.

# Interested parties

@ruebot

### DIFF
--- a/src/main/java/io/archivesunleashed/data/ArcRecordUtils.java
+++ b/src/main/java/io/archivesunleashed/data/ArcRecordUtils.java
@@ -34,8 +34,13 @@ import org.archive.io.arc.ARCRecordMetaData;
  * Utilities for working with {@code ARCRecord}s (from archive.org APIs).
  */
 public final class ArcRecordUtils {
-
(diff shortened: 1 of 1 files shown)


ianmilligan1:
Oh just saw #135 come in, which is a better solution than this!

Issue #124, opened 2017-11-30
NullPointerException error during build


ruebot:
```

Running io.archivesunleashed.spark.matchbox.ComputeImageSizeTest
java.lang.NullPointerException
	at java.io.ByteArrayInputStream.<init>(ByteArrayInputStream.java:106)
	at io.archivesunleashed.spark.matchbox.ComputeImageSize$.apply(ComputeImageSize.scala:28)
	at io.archivesunleashed.spark.matchbox.ComputeImageSizeTest$$anonfun$1.apply(ComputeImageSizeTest.scala:39)
	at io.archivesunleashed.spark.matchbox.ComputeImageSizeTest$$anonfun$1.apply(ComputeImageSizeTest.scala:36)
	at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
	at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
	at org.scalatest.Transformer.apply(Transformer.scala:22)
	at org.scalatest.Transformer.apply(Transformer.scala:20)
	at org.scalatest.FunSuiteLike$$anon$1.apply(FunSuiteLike.scala:186)
	at org.scalatest.TestSuite$class.withFixture(TestSuite.scala:196)
	at org.scalatest.FunSuite.withFixture(FunSuite.scala:1560)
	at org.scalatest.FunSuiteLike$class.invokeWithFixture$1(FunSuiteLike.scala:183)
	at org.scalatest.FunSuiteLike$$anonfun$runTest$1.apply(FunSuiteLike.scala:196)
	at org.scalatest.FunSuiteLike$$anonfun$runTest$1.apply(FunSuiteLike.scala:196)
	at org.scalatest.SuperEngine.runTestImpl(Engine.scala:289)
	at org.scalatest.FunSuiteLike$class.runTest(FunSuiteLike.scala:196)
	at org.scalatest.FunSuite.runTest(FunSuite.scala:1560)
	at org.scalatest.FunSuiteLike$$anonfun$runTests$1.apply(FunSuiteLike.scala:229)
	at org.scalatest.FunSuiteLike$$anonfun$runTests$1.apply(FunSuiteLike.scala:229)
	at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:396)
	at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:384)
[...]

PR #141, opened 2017-12-06
Remove stackTrace output on caught error.

greebie:
**Remove Stack Trace Output on an Expected Error**

* * *

**GitHub issue(s)**:

#124 

Unit tests were producing a "NullPointerError" that was caught and then thrown during the testing period.  

Since the try-catch call is both expected on a ComputeImageSize(null) and returns (0,0) regardless, there is not reason to output the stack trace here.

# How should this be tested?

Compilation does not print out NullPointer error anymore.

# Additional Notes:
N/A

# Interested parties

### DIFF
--- a/src/main/scala/io/archivesunleashed/spark/matchbox/ComputeImageSize.scala
+++ b/src/main/scala/io/archivesunleashed/spark/matchbox/ComputeImageSize.scala
@@ -32,7 +32,6 @@ object ComputeImageSize {
       (image.getWidth(), image.getHeight())
     } catch {
       case e: Throwable => {
-        e.printStackTrace()
         return (0, 0)
       }
     }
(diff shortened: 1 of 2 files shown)


greebie:
The error is thrown when ComputeImageSizeTest purposely (and rightfully) calls ComputeImageSize(null) and expects a return of (0,0).  

The error gets caught by ComputeImageSize who in turn prints the error (needlessly imho) to the screen.  My review of similar tests suggests that printing to screen is not necessary in this particular case.

tl;dr the error you are seeing is expected by the test.  It's just being seen by output due to the printout.

greebie:
Thanks.  Worth keeping this Issue in mind.  The only thing I can think of might be some form of crazy IO error.  In this case, the code will return (0,0) and move on, which is the correct path to take.  The only risk is the very rare case that a crazy IO error causes a different output than usual.  But that's going to be a) quite rare and b) negligible for any large sized dataset.

greebie:
Walking to my car, I wonder if it's worth sending an error warning instead of the stack trace.  I'll mull it over.

PR #150, opened 2017-12-11
Add more packages and exclusions for #113

ruebot:
**GitHub issue(s)**: 

* #113  
* #119
* https://github.com/archivesunleashed/aut/issues/113#issuecomment-350675138

Adds some more explicit dependencies to `pom.xml` and another exclusion so we can run spark-shell with `--packages` instead of `--jars`.

# How should this be tested?

I've tested it [here](https://gist.github.com/ruebot/bd6a6b2993d21171eda839725cc66438) with a local branch of docker-aut. You can do the same if you like with this [`Dockerfile`](https://gist.github.com/ruebot/bf5c625b9edb256e8fe4286615ebf7ce).

# Interested parties

@dportabella @ianmilligan1 @lintool @greebie

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -454,6 +454,12 @@
       <groupId>org.scalanlp</groupId>
       <artifactId>breeze_2.11</artifactId>
       <version>0.13.1</version>
+      <exclusions>
+        <exclusion>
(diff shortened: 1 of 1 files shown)


ianmilligan1:
👍 I'll test locally but looks good to me!

PR #152, opened 2017-12-11
Too many JUNITs

ruebot:
Follow-up to #150 

@ianmilligan1


ianmilligan1:
Aye (sorry for delay, daycare ended!)

Issue #154, opened 2017-12-15
ARC Handling Bug in 0.12.0 when Extracting Links


ianmilligan1:
This bug is confirmed for me on aut 0.12.0, but when I rolled back to 0.10.0, the script worked.

Running following script on Ubuntu 16 AWS machine.

```
import io.archivesunleashed.spark.matchbox.{ExtractDomain, ExtractLinks, RecordLoader}
import io.archivesunleashed.spark.rdd.RecordRDD._
val links = RecordLoader.loadArchives("/data/cpp/*.arc.gz", sc).map(r => (r.getCrawlDate, ExtractLinks(r.getUrl, r.getContentString))).flatMap(r => r._2.map(f => (r._1, ExtractDomain(f._1).replaceAll("^\\s*www\\.", ""), ExtractDomain(f._2).replaceAll("^\\s*www\\.", "")))).filter(r => r._2 != "" && r._3 != "").countItems().filter(r => r._2 > 5).saveAsTextFile("/data/all-links")
```
On aut 0.10.0 it works. 

On aut 0.12.0, I get the following error. [Full Gist here](https://gist.github.com/ianmilligan1/babd67e78638495bba9f9fc36eb30b5f).

```
[Stage 0:>                                                      (19 + 8) / 2441]2017-12-15 16:32:17,654 [Executor task launch worker for task 26] ERROR Executor - Exception in task 26.0 in stage 0.0 (TID 26)
java.lang.StackOverflowError
	at org.jsoup.parser.HtmlTreeBuilderState$7.process(HtmlTreeBuilderState.java:279)
	at org.jsoup.parser.HtmlTreeBuilder.process(HtmlTreeBuilder.java:142)
	at org.jsoup.parser.HtmlTreeBuilderState$15.anythingElse(HtmlTreeBuilderState.java:1207)
	at org.jsoup.parser.HtmlTreeBuilderState$15.process(HtmlTreeBuilderState.java:1201)
	at org.jsoup.parser.HtmlTreeBuilder.process(HtmlTreeBuilder.java:137)
	at org.jsoup.parser.TreeBuilder.processStartTag(TreeBuilder.java:66)
	at org.jsoup.parser.HtmlTreeBuilderState$7.process(HtmlTreeBuilderState.java:707)
	at org.jsoup.parser.HtmlTreeBuilder.process(HtmlTreeBuilder.java:142)
[...]

PR #155, opened 2017-12-15
Changes jsoup version in pom.xml (#154)

ianmilligan1:
As seen in [Issue 154](https://github.com/archivesunleashed/aut/issues/154), the changed jsoup version seems to have blown up link extraction in some ARC files. 

I've verified on multiple systems, and with version 1.7.3 it works and on version 1.11.1 it fails, at least with the Canadian labour collection.

That said, we changed to version 1.11.1 to resolve [this issue](https://github.com/archivesunleashed/aut/issues/113), so I'm not sure what the best way to proceed is. I am also very inexperienced in dependency issues, so am completely deferring here.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -483,7 +483,7 @@
     <dependency>
       <groupId>org.jsoup</groupId>
       <artifactId>jsoup</artifactId>
-      <version>1.11.1</version>
+      <version>1.7.3</version>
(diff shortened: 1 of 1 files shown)


ianmilligan1:
Heh yeah, Mr. Ubuntu here. 😃

PR #161, opened 2018-01-06
Changing keepDate to allow multiple dates, would close #108

### LINKED ISSUE #108: KeepDate UDF should support date range
Right now, our filter by date operates like so:

`.keepDate("200810", YYYYMM)` (returning October 2008 hits)

So if we say wanted September, October, and November, we'd have to layer commands like so:

```
.keepDate("200809", YYYYMM)
.keepDate("200810", YYYYMM)
.keepDate("200811", YYYYMM)
```

It'd be nice to be able to pass a date range, i.e. `.keepDate("200809","200811",YYYYMM`

ianmilligan1:
**GitHub issue(s)**:

* #108

This Pull Request changes the use of `keepDate`. Right now, it is used like: 

```
  .keepDate("2008", YYYY)
```

If a user wanted data from 2008 and 2010, they would have to run two separate scripts - one for 2008 and one for 2010.

This new format allows the use of a list in `keepDate`, so a command like:

```
  .keepDate(List("2007","2010"), YYYY)
```

# How should this be tested?

I have tested this on a collection of WARCs with files from 2007 and from 2010. Running:

```
import io.archivesunleashed.spark.matchbox.RecordLoader
import io.archivesunleashed.spark.rdd.RecordRDD._
import io.archivesunleashed.spark.matchbox.{RemoveHTML, RecordLoader}
import io.archivesunleashed.spark.matchbox.ExtractDate.DateComponent._

RecordLoader.loadArchives("/Users/ianmilligan1/dropbox/testdata/*.gz", sc)
  .keepValidPages()
  .keepDate(List("2007","2010"), YYYY)
  .map(r => (r.getCrawlDate, r.getDomain, r.getUrl, RemoveHTML(r.getContentString)))
  .saveAsTextFile("/users/ianmilligan1/desktop/output/2007and2010")
```

Resulted in all the records from 2007 and 2010. You can also still test only one year, but it still needs to be provided as a list.

# Additional Notes:

If approved, this will require revision to all `keepDate` documentation.

# Interested parties

@ruebot @greebie @lintool

### DIFF
--- a/src/test/scala/io/archivesunleashed/spark/ArcTest.scala
+++ b/src/test/scala/io/archivesunleashed/spark/ArcTest.scala
@@ -45,12 +45,12 @@ class ArcTest extends FunSuite with BeforeAndAfter {
 
   test("filter date") {
     val four = RecordLoader.loadArchives(arcPath, sc, keepValidPages = false)
-      .keepDate(List("200804"), DateComponent.YYYYMM)
+      .keepDate(List("200804","200805"), DateComponent.YYYYMM)
       .map(r => r.getCrawlDate)
       .collect()
 
     val five = RecordLoader.loadArchives(arcPath, sc, keepValidPages = false)
-      .keepDate(List("200805"), DateComponent.YYYYMM)
+      .keepDate(List("200805","200807"), DateComponent.YYYYMM)
       .map(r => r.getCrawlDate)
       .collect()
 
--- a/src/test/scala/io/archivesunleashed/spark/ArcTest.scala
+++ b/src/test/scala/io/archivesunleashed/spark/ArcTest.scala
@@ -45,12 +45,12 @@ class ArcTest extends FunSuite with BeforeAndAfter {
 
   test("filter date") {
     val four = RecordLoader.loadArchives(arcPath, sc, keepValidPages = false)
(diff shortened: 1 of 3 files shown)


ianmilligan1:
Hmm, the code coverage is slightly lower, but so far as I can tell, that line was never covered before. Maybe just because it is longer now?

greebie:
Yes.  If you add content to an uncovered line, it will decrease overall coverage.

src/test/scala/io/archivesunleashed/spark/ArcTest.scala
```diff
@@ -45,12 +45,12 @@ class ArcTest extends FunSuite with BeforeAndAfter {
 
   test("filter date") {
     val four = RecordLoader.loadArchives(arcPath, sc, keepValidPages = false)
-      .keepDate("200804", DateComponent.YYYYMM)
+      .keepDate(List("200804"), DateComponent.YYYYMM)
```
ruebot: In your example you use two dates, you should use two dates here as well.

ianmilligan1: Added in latest commit, thanks for the review!

PR #167, opened 2018-01-28
Graphml

greebie:
**FIX GraphML Output**

* * *

**GitHub issue(s)**:

NA

Fixes problem with label ids.  Unfortunately, this will downgrade the graphml to what we had with GML, where node labels must be copied to gephi by default from the id.

# How should this be tested?

It should work with a) Gephi and b) Borg-reducer.

# Additional Notes:

Future development should attempt to produce a numeric id for nodes and labels.  The possible best way forward is to produce a hash based on the url strings, in order to make it unique.

# Interested parties

@ianmilligan1 


ianmilligan1:
Looks good!

> Unfortunately, this will downgrade the graphml to what we had with GML, where node labels must be copied to gephi by default from the id

In my gephi, the node labels show up perfectly (see screenshot) with the GraphML generated using this. 

When we use borg-reducer, it creates an `id` and an `Id` in the file.. I'll open up an issue there.

PR #175, opened 2018-03-07
make ArchiveRecord a trait

helgeho:
**The title of this pull-request should be a brief description of what the pull-request fixes/improves/changes. Ideally 50 characters or less.**

* * *

**GitHub issue(s)**:

* ex: #1
* ex: #2

A brief description of what the intended result of the Pull Request (PR) will be, what problem it solves, technical details, and any possible side effects.

Example:
* Changes x feature to such that y
* Added x
* Removed y

# How should this be tested?

A description of what steps someone could take to:
* Reproduce the problem you are fixing (if applicable)
* Test that the PR does what is intended.
* Please be as detailed as possible.
* Good testing instructions help get your PR completed faster.

# Additional Notes:

Any additional information that you think would be helpful when reviewing this PR.

Example:
* Does this change require documentation to be updated?
* Does this change add any new dependencies?
* Could this change or impact execution of existing code?

# Interested parties

### DIFF
--- a/src/main/scala/io/archivesunleashed/spark/archive/io/ArchiveRecord.scala
+++ b/src/main/scala/io/archivesunleashed/spark/archive/io/ArchiveRecord.scala
@@ -29,7 +29,31 @@ import io.archivesunleashed.io.ArchiveRecordWritable.ArchiveFormat
 import io.archivesunleashed.spark.matchbox.ExtractDate.DateComponent
 import io.archivesunleashed.spark.matchbox.{RemoveHttpHeader, ExtractDate, ExtractDomain}
 
-class ArchiveRecord(r: SerializableWritable[ArchiveRecordWritable]) extends Serializable {
+object ArchiveRecord {
+  val ISO8601 = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssX")
+}
+
+trait ArchiveRecord extends Serializable {
+  def getCrawlDate: String
+
+  def getCrawlMonth: String
+
+  def getContentBytes: Array[Byte]
+
+  def getContentString: String
+
+  def getMimeType: String
+
+  def getUrl: String
+
+  def getDomain: String
+
+  def getImageBytes: Array[Byte]
+}
+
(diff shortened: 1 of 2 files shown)


greebie:
I'll give this a test run today.  Overall it lgtm - it's just a matter of ensuring the RecordLoader works the same as last time (and based on the code, I see no reason why it should not).

greebie:
Confirmed lgtm.  Tested with plain text extraction and WriteGEXF.  RecordLoader works as expected (as expected).

greebie:
Tested with both arcs and warcs as well.

lintool:
Taking a look at this - @helgeho what's the rationale for both an object `ArchiveRecord` and a trait `ArchiveRecord`?

cjer:
AFAIK it's good practice to put (what would be in other languages) "static" methods (e.g. `ISO8601`) in a *companion object*, which is an object with the same name as the class or trait. It also has higher access privileges to that trait than other objects.

https://alvinalexander.com/scala/how-to-static-members-in-scala-companion-objects-fields-methods

lintool:
Thanks for the info. In which case something like this should documented.

greebie:
Flagging this to issue #176 for when I go through code documentation.

helgeho:
@lintool, sorry about the confusion, but @cjer is correct, and that's why in my opinion this does *not* need to be documented, instead, anything that's not considered good practice should be documented, that's what we have design patterns and conventions for...

lintool:
@helgeho I see.

Nevertheless, the rationale for turning `ArchiveRecord` into a trait should have been documented - for example, the PR template was never even filled out.

Plus, impl should really have been moved into it's own file.

helgeho:
I asked @ruebot before, whether it would be okay to do so as the previous implementation was very inflexible with the `val`s instead of `def`s, I needed this flexibility for https://github.com/helgeho/ArchiveSpark-AUT-bridge. I kind of disagree with the own file, since this is the standard implementation and it's not java, but feel free to refactor that of course 🙂 I'll fill out the PR template next week as I am currently on vacation (writing this from my phone). Best, Helge

PR #186, opened 2018-03-27
make ArchiveRecord a trait

helgeho:
**GitHub issue(s)**:

* https://github.com/archivesunleashed/aut/pull/175
* https://github.com/archivesunleashed/aut/pull/181

This commit introduces a trait that describes the interface of an `ArchiveRecord` in AUT. This allows for alternative implementations of this core record type in AUT, such as the one provided by [ArchiveSpark-AUT-bridge](https://github.com/helgeho/ArchiveSpark-AUT-bridge). The original AUT implementation of `ArchiveRecord`, which is instantiated by AUT's `RecordLoader`, is moved to the `ArchiveRecordImpl` class.

This was already done before by https://github.com/archivesunleashed/aut/pull/175 and reverted by https://github.com/archivesunleashed/aut/pull/181 because the change of moving `ISO8601` into a *static object* caused some issues with certain collections (see below). Therefore, this time, `ISO8601` remains part of `ArchiveRecordImpl`.

# How should this be tested?

As this is a pure refactoring without changing any logic, a simple compilation + smoke test should be sufficient. However, as it turns out, the little additional change of moving `ISO8601` into a *static object* in the first attempt of this change (https://github.com/archivesunleashed/aut/pull/175) caused strange behaviors with some collections (probably when running on multiple cores per executor, since `SimpleDateFormat` is not thread-safe as pointed out by @anjackson, see https://github.com/archivesunleashed/aut/pull/181). Hence, it should be better tested with different kinds of collections to make sure there are no unexpected side effects involved.

# Additional Notes:

The first attempt to make ArchiveRecord a trait (https://github.com/archivesunleashed/aut/pull/175) was reverted by https://github.com/archivesunleashed/aut/pull/181. This time nothing else is changed except for the trait (`ISO8601` remains part of `ArchiveRecordImpl`). @ruebot and @ianmilligan1 please test it as you did last time (https://github.com/archivesunleashed/aut/pull/181), I hope it is not causing problems this time.


ianmilligan1:
Thanks @helgeho - I should be able to kick the tires on this later this week (or early next in worst case - it's a tough time of term).

src/main/scala/io/archivesunleashed/spark/archive/io/ArchiveRecordImpl.scala
```diff
@@ -0,0 +1,96 @@
...
+      arcRecord.getMetaData.getUrl
+    } else {
+      warcRecord.getHeader.getUrl
+    }
+  }
+
+  val getDomain: String = ExtractDomain(getUrl)
+
+  val getImageBytes: Array[Byte] = {
+    if (getContentString.startsWith("HTTP/"))
```
lintool: for consistency, add braces also?

helgeho: I didn't change anything there, that's exactly your current https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/spark/archive/io/ArchiveRecord.scala , I only wanted to add the trait, not make your code more / less consistent, because I do not know how your policy is in terms of coding styles (personally though, I also prefer no braces for one-liners)

helgeho:
https://github.com/helgeho/ArchiveSpark-AUT-bridge is also updated now, please check it out, give it a try!

lintool:
Neat! Although you'll want to change your documentation to reflect the new package structure.

helgeho:
oh, that's right, thanks for the heads up, I'll do that soon

Issue #220, opened 2018-05-14
DataFrames for image analysis


lintool:
Currently, we have RDD-based analytics for image analysis here:
https://archivesunleashed.org/aut/#image-analysis

Let's DataFrame-ify it - that is, build the DataFrame infrastructure that wold support image analysis.

I'm thinking of creating two separate DataFrames:
+ The first for the image links, something like (source page, image url)
+ The second for the images themselves, something like (img url, type, height, width, md5, raw bytes)

JWZ2018:
I will work on this.

lintool:
@JWZ2018 

Great. BTW - one end game would be to create something like this in a single e2e pipeline: http://ruebot.net/elxn42.html

Fork the repo, start working on a branch - send an initial PR when you're ready and we can discuss iteratively.

ruebot:
[Here's](http://ruebot.net/2016/03/1203867-elxn42-images/) the background on how I make those.

JWZ2018:
@lintool 
Just making sure I'm on the right track and understanding the codebase properly, for the above two dataframes, I'm planning to add `extractImageLinksDF` and `extractImageDetailsDF` to the `WARecordRDD` in `io/archivesunleashed/package.scala` and then add them to `DataframeLoader` similar to how `extractValidPagesDF` and `extractHyperlinksDF` are implemented.
Does that sound right?

lintool:
Yes, that's a good start!

JWZ2018:
@lintool 
Clarifying for the second df:
- type means png/jpg?
- what is a good way to find the height and width of the image? I did not find it in the WARCRecord or the Warc file format from https://iipc.github.io/warc-specifications/specifications/warc-format/warc-1.0/. Maybe it can be computed from the raw bytes?

lintool:
With #226 this is done. Closing.

PR #221, opened 2018-05-15
Extract Image Links DF API + Test

JWZ2018:
* * *

**GitHub issue(s)**:

* #220

* Add DataFrame API to extract (source page, image url) pairs in `WARRecord`
* Add an entry point from DataframeLoader
* Add a test for the API

# How should this be tested?

* `mvn clean install`
* `mvn -Dtest=ExtractImageLinksTest test`

# Additional Notes:
* `mvn clean install` builds successfully and all tests pass
* I could not get the local snapshot build to run in the spark-shell. When I run `spark-shell --packages "io.archivesunleashed:aut:0.16.1-SNAPSHOT"`, I get 
screen shot 2018-05-14 at 10 34 36 pm
Not sure if I'm starting the shell correctly for a snapshot build.

# Interested parties
@lintool @ruebot

### DIFF
--- a/src/main/scala/io/archivesunleashed/DataFrameLoader.scala
+++ b/src/main/scala/io/archivesunleashed/DataFrameLoader.scala
@@ -14,6 +14,7 @@ class DataFrameLoader(sc: SparkContext) {
       .extractHyperlinksDF()
   }
 
+  /* Create a dataframe with (source page, image url) pairs */
   def extractImageLinks(path: String): DataFrame = {
   	RecordLoader.loadArchives(path, sc)
   		.extractImageLinksDF()
--- a/src/main/scala/io/archivesunleashed/DataFrameLoader.scala
+++ b/src/main/scala/io/archivesunleashed/DataFrameLoader.scala
@@ -13,4 +13,9 @@ class DataFrameLoader(sc: SparkContext) {
     RecordLoader.loadArchives(path, sc)
       .extractHyperlinksDF()
   }
+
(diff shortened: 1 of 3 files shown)


src/main/scala/io/archivesunleashed/DataFrameLoader.scala
```diff
@@ -13,4 +13,9 @@ class DataFrameLoader(sc: SparkContext) {
     RecordLoader.loadArchives(path, sc)
       .extractHyperlinksDF()
   }
+
```
ruebot: Add a doc comment here.

src/main/scala/io/archivesunleashed/package.scala
```diff
@@ -120,6 +120,24 @@ package object archivesunleashed {
       sqlContext.getOrCreate().createDataFrame(records, schema)
     }
 
+    def extractImageLinksDF(): DataFrame = {
```
ruebot: Need a doc comment.

lintool:
lg so far.

Anything obvious that we should add to this DF? Maybe the alt text?

Let's get the DF for the images themselves, then we can do interesting analyses like, what's the most common images that are linked to. With de-dup'ing on MD5, we can look at cases where the same exact images are being copied around and renamed.

JWZ2018:
I will add the DF for the actual images in a separate PR to not clutter this one.

lintool:
Okay, make the edits by @ruebot and I'll give a +1.

lintool:
lgtm, go ahead and merge please.

PR #224, opened 2018-05-15
Remove duplicate call of keepValidPages

jwli229:

As title

# How should this be tested?

* `mvn clean install`

# Interested parties

* @lintool 
* @ruebot


src/main/scala/io/archivesunleashed/package.scala
```diff
@@ -105,7 +105,7 @@ package object archivesunleashed {
     }
 
     def extractHyperlinksDF(): DataFrame = {
-      val records = rdd.keepValidPages()
+      val records = rdd
```
ruebot: @lintool do you have a preference of where it is removed? I'm good with how it is.

lintool:
codecov is really wacked sometimes.... why red if there's a 0% diff in coverage?

PR #230, opened 2018-05-21
Code cleanup: ArchiveRecord + impl moved into same Scala file

lintool:

Cleans up code for `ArchiveRecord` and adds documentation.

+ Per https://docs.scala-lang.org/style/control-structures.html - Omit braces if you have an `else` clause.
+ Per https://docs.scala-lang.org/style/files.html - As a rule, files should contain a single logical compilation unit. By “logical” I mean a class, trait or object.
+ Added documentation.


src/main/scala/io/archivesunleashed/ArchiveRecord.scala
```diff
@@ -17,20 +17,111 @@
 
 package io.archivesunleashed
 
+import java.text.SimpleDateFormat
+
```
ruebot: Weird. My comment from earlier is missing.

...do we need this blank line?

lintool: I can't find Scala style guides that dictate import, but for Java, it is common to have groups separated by blank lines, see http://checkstyle.sourceforge.net/config_imports.html

To quote, one common approach is:

+ group of static imports is on the top
+ groups of non-static imports: "java" and "javax" packages first, then "org" and then all other imports
+ imports will be sorted in the groups
+ groups are separated by, at least, one blank line

ruebot: Ah, ok. That make sense. Let's just proceed with using that approach.

As an aside, I had the same problem trying to find guidelines on Scala style imports, and ended up doing it alphabetically, and soon realized order does matter at times. :man_shrugging:

Issue #203, opened 2018-04-26
Refactor ExtractGraph and assess value of GraphX for producing network graphs


greebie:
The upgrade to 2.3 fails ExtractGraphTest, but that test and the associated udf are under-utilized and could use some refactoring.

Let's use this opportunity to re-examine the value of ExtractGraph into the wasapi - aut - graphpass - auk pipeline.

ruebot:
Context: The potential upgrade to Spark 2.3.0 in the issue-197 branch (#197).

lintool:
As a first task, let's try [to extract a domain link graph](http://archivesunleashed.org/aut/#analysis-of-site-link-structure) and convert it into a GraphX graph object - and from there we can get access to GraphX features...

greebie:
Note that Graphx (I called it Sparkx and that is wrong) is not compatible with Dataframes.  GraphFrames may be an alternative:  http://graphframes.github.io/

ianmilligan1:
>Note that Graphx (I called it Sparkx and that is wrong) is not compatible with Dataframes. GraphFrames may be an alternative: http://graphframes.github.io/

Given that GraphX is not comparable with the future direction of the project (Dataframes), should we close this issue?

greebie:
@hardiksahl 's work provides some helpful new direction, but it will not resolve this issue in the short-run.

greebie:
After looking at the issue, there is another way forward, using connected components and graphpass.

If we use GraphX instead of the usual, we will have access Page Rank values, connected components and strongly connected components.  

In Graphpass, if the the node list > 50,000, then we can try filtering using just the largest connected component (and the data is all ready there, so it is not a huge run).

If the node list is still > 50,000 then we can use the graph of just the strongly connected components.  In this case, it means that we have the "true core" of websites in interaction with each other.  It is not a full representation of the graph, but it is a logical criteria for a "first pass" in the graph.

greebie:
In other news, as a general feature, it may be possible to offer a Gexf that at least provides colour (using components) and size visualization features for a Gephi output.  Positioning will still be random, but if someone does not want to use GraphPass, it's still better than the status quo.

ruebot:
Can we rewrite this issue, or close and open up some news ones? It seems like there are a few different things going on here, and it's not exactly clear what the definition of done is.

ianmilligan1:
[...]

PR #245, opened 2018-07-26
Add ExtractGraphX including algorithms for PageRank and Components. Issue 203

greebie:
**GitHub issue(s)**:

#203

Adds ExtractGraphX algorithm and GraphML output to go with GraphX output.
Adds feature to calculate pagerank and weak and strong component calculations.
Provides some lint fixes for other files (usually removing Magic Numbers)
Deprecates ExtractGraph as outdated (although there is some discussion that comes with that).

# How should this be tested?

```
import io.archivesunleashed._
import io.archivesunleashed.util._
import org.apache.spark.graphx._

val graph = ExtractGraphX.extractGraphX(RecordLoader.loadArchives("/Users/USERNAME/WARCFOLDER/", sc)
.keepValidPages()
.flatMap(r => ExtractLinks(r.getUrl, r.getContentString))
.map(r => (ExtractDomain(r._1).removePrefixWWW(), ExtractDomain(r._2).removePrefixWWW()))
.filter(r => r._1 != "" && r._2 != "")).subgraph(epred = eTriplet => eTriplet.attr.edgeCount>5)

val pRank = ExtractGraphX.runPageRankAlgorithm(graph)

WriteGraphXML(pRank, "graphML-path.graphml/")
```

# Additional Notes:

The main feature is `strong` and `weak` connected components which can be used in Graphpass to reduce the size of a network graph if it is > 50k nodes.
`strong` connected components are also interesting as a potential measure of who is driving the overall conversation, or if factions exist in a community.

# Interested parties


greebie:
ExtractGraph does provide a rudimentary json output that is not included in ExtractGraphX.  I think it makes sense to create a more generic CreateJSONGraph or even a more detailed output class that will do whatever format we want. Either way, I'm not sure if the json creator is used very much.

ianmilligan1:
Thanks for this Ryan!

For the test script, what should we expect to see as a result?

greebie:
It should produce a network graph at "graphML-path.graphml/" in graphml format with pagerank and other metadata.

ianmilligan1:
OK great, thanks @greebie - I will test it, probably tomorrow morning!

ianmilligan1:
Successfully generated the file with:

```
import io.archivesunleashed._
import io.archivesunleashed.util._
import org.apache.spark.graphx._

val graph = ExtractGraphX.extractGraphX(RecordLoader.loadArchives("/mnt/vol1/data_sets/cpp/cpp_warcs_accession_01/partner.archive-it.org/cgi-bin/getarcs.pl/*200912*", sc)
.keepValidPages()
.flatMap(r => ExtractLinks(r.getUrl, r.getContentString))
.map(r => (ExtractDomain(r._1).removePrefixWWW(), ExtractDomain(r._2).removePrefixWWW()))
.filter(r => r._1 != "" && r._2 != "")).subgraph(epred = eTriplet => eTriplet.attr.edgeCount>5)

val pRank = ExtractGraphX.runPageRankAlgorithm(graph)

WriteGraphXML(pRank, "/mnt/vol1/derivative_data/test/graphML-test.graphml")
```

However, when I attempted to open the ensuing `graphML-test.graphml` file in Gephi, received the following error message:

```
javax.xml.stream.XMLStreamException: ParseError at [row,col]:[10,51]
Message: Element type "key" must be followed by either attribute specifications, ">" or "/>".
	at com.sun.org.apache.xerces.internal.impl.XMLStreamReaderImpl.next(XMLStreamReaderImpl.java:604)
	at org.gephi.io.importer.plugin.file.ImporterGraphML.execute(ImporterGraphML.java:158)
Caused: java.lang.RuntimeException
	at org.gephi.io.importer.plugin.file.ImporterGraphML.execute(ImporterGraphML.java:181)
	at org.gephi.io.importer.impl.ImportControllerImpl.importFile(ImportControllerImpl.java:199)
	at org.gephi.io.importer.impl.ImportControllerImpl.importFile(ImportControllerImpl.java:169)
	at org.gephi.desktop.importer.DesktopImportControllerUI$4.run(DesktopImportControllerUI.java:341)
Caused: java.lang.RuntimeException
	at org.gephi.desktop.importer.DesktopImportControllerUI$4.run(DesktopImportControllerUI.java:349)
[catch] at org.gephi.utils.longtask.api.LongTaskExecutor$RunningLongTask.run(LongTaskExecutor.java:274)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

greebie:
Re: auk.

The first thing is that this update offers some good features for network analysis, so it is worthwhile adding, even if it does not work for AUK.

The AUK spark script would need revision, which is mostly wrapping the current flatMap script in the GraphX graph object and using .runPageRankAlgorithm(). It will afford an additional advantage that the raw Gephi graphs could also be slightly more attractive, because they use pagerank instead of degree as the default sizing.

I'll definitely do some testing before we move forward on changing the algorithm. The new approach will take generally longer for all graphs, but much much shorter than what GraphPass does for huge ones.

Also, Hardik and I were looking at a paper that examined the differences in pageRank calculations from Spark to Igraph.

One downside is that the spark run would take longer. So we would need to take a cost-benefit and/or consider whether its worthwhile to have some conditionals before running this approach.  

The "quickrun" feature in GraphPass would include something like this pseudocode:

```
i f NODESIZE (graph) > 50 0000:
  check for "Weak" and/or "Strong" attribute
    create the subgraph from largest weak components if the attr exists and check if the new graph < 50000 nodes
    otherwise
    create the strong subgraph if the attr exists and check if the new graph < 50000 nodes
    otherwise fail graphpass.
```

   if we have a good new graph, then run the usual quickpass stuff.
   maybe send a _WEAKPASS or _STRONGPASS file to tell AUT we didn't use the whole graph.

The downside is longer Spark calls, possibly for all graphs if we do not add some conditional reasoning behind choosing to run the PageRank algorithm
 
The upside is that this would be able to provide visualizations for very large graphs in a way that makes theoretical sense.  Since the web archives are not going to get smaller in the long term, it is important we have a solution, even if its not optimal.

Further, we have a way forward if SNAP is simply beyond my capabilities or does not have the ability to do what we need it to.

Graphpass would not to eat up resources for huge graphs.

greebie:
@ianmilligan1 Yup.  That looks like I accidentally added or removed a "<" somewhere.  Will revise.

greebie:
@ianmilligan1 The latest update should work in Gephi this time.  I tested it with a pretty good set of WARCs.  The weak component works nicely!

greebie:
Hey 70% Codecov!  That's a milestone!  :)

src/main/scala/io/archivesunleashed/ArchiveRecord.scala
```diff
@@ -98,14 +98,14 @@ class ArchiveRecordImpl(r: SerializableWritable[ArchiveRecordWritable]) extends
     new String(getContentBytes)
   }
 
-  val getMimeType = {
+  val getMimeType: String = {
```
ianmilligan1: Just a quick question - what's this change for?

Issue #212, opened 2018-04-27
Convert nulls to Option(T)


greebie:
Scala recommends not using `null` from Java objects and instead using `Option(T)`. The advantage is to converting is that it will avoid `NullPointerError` in the code which can be a pain especially in huge files.

I don't think this is a high priority issue, but it may clear up some errors down the road.

This was discovered during the fix in #196.

greebie:
In general, the approach here is:
- all checks for `(x != null)` are changed to `(Option(x) != None)`
- data checks use 
````
val maybeX: Option[String/Int/Whatever] = Option(x)
maybeX match {
  case Some(x) =>
    // do the usual stuff
  case None =>
    // default value or do nothing
}
```
- the default value will be settable by client. It would be good to have a generic global constant for default returns on empty strings in aut.  So far I'm using "aut_empty."
- I think we should remove try-catch in basic udfs. The Option should handle most things that matter. If there is an IOError, that should be caught at the RecordLoader stage, or we will get a TypeError that will blow up before we get to the udf.

PR #249, opened 2018-08-07
Deal with final scalastyle assessments: Issue 212

greebie:
**GitHub issue(s)**:

#212

It finishes off the scalastyle warnings by dealing with a number of `null` calls which scalastyle does not like.

image

Provides the final revisions to scalastyle problems, sometimes resulting in code refactoring.

In some cases, I felt like the Option[t] approach simplified the code and so I refactored (examples include ExtractDomain and ExtractLinks). These also had other style errors like using `return.`

I tried not to go overboard doing this, however.

In other cases, the only thing needed was to return an empty string "" instead of null.

In some cases, there seemed like no point in worrying about null. In this case, I used `// scalastyle:off null` Example is the NER stuff. 

For tests, the nulls were kept as a way to break things.  I used `scalastyle:off null` there as well.

Overall, whether we like the Option[t] approach or not, I think having the scalastyle stuff formalized and clean will be good for future.  We can also just turn off the "null" setting in the configuration.

# How should this be tested?

Because this involved some code refactoring of pretty major functions (ExtractDomain e.g.) I would suggest testing it with a messy set of warcs just to make sure I did not break things.  For the most part, other than switching "nulls" to empty strings, all of the current tests are the same and passed.

For discussion, is the idea of outputting a default bad output string: something like "aut_empty" that could be set globally. However, for now I just went with empty string.

# Additional Notes:

I made it so that using the Option[t] did not affect example scripts.  In some ways it would be better for the functions to send out Option[t]s but it would mean that people would have to learn scala in more detail which is something we are pulling away from.

# Interested parties

@ianmilligan1 @lintool (because of style issues raised) @ruebot 

### DIFF
--- a/src/main/scala/io/archivesunleashed/matchbox/ExtractBoilerpipeText.scala
+++ b/src/main/scala/io/archivesunleashed/matchbox/ExtractBoilerpipeText.scala
@@ -29,15 +29,12 @@ object ExtractBoilerpipeText {
    * @return text with boilerplate removed or Nil if the text is empty.
    */
   def apply(input: String): String = {
-    try {
-      if (input.isEmpty) {
-        null
-      } else {
+    val maybeInput: Option[String] = Option(input)
+    maybeInput match {
+      case Some(input) =>
         extract(input)
-      }
-    } catch {
-      case e: Exception =>
-        throw new IOException("Caught exception processing input row " + e)
+      case None =>
+        ""
     }
   }
 
@@ -47,11 +44,13 @@ object ExtractBoilerpipeText {
    * @return filtered text or Nil if the text is empty.
    */
   def extract (input: String): String = {
-    val text = DefaultExtractor.INSTANCE.getText(input).replaceAll("[\\r\\n]+", " ").trim()
-    if (text.isEmpty) {
-      null
-    } else {
-      text
+    val maybeText: Option[String] = Option(DefaultExtractor.INSTANCE
+      .getText(input).replaceAll("[\\r\\n]+", " ").trim())
+    maybeText match {
+      case Some(text) =>
+        text
+      case None =>
+        ""
     }
   }
 }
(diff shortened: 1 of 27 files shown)


src/main/scala/io/archivesunleashed/matchbox/ExtractBoilerpipeText.scala
```diff
@@ -29,29 +29,13 @@ object ExtractBoilerpipeText {
...
-      } else {
-        extract(input)
-      }
-    } catch {
-      case e: Exception =>
-        throw new IOException("Caught exception processing input row " + e)
-    }
-  }
-
-  /** Extracts boilerplate.
```
ruebot: Why is this function being removed?

ianmilligan1: Let's keep `ExtractBoilerpipeText` - it's one of our core functions, is documented, and is out of scope for this PR in any case (I suspect its removal is accidental?).

greebie: I believe the code on the right does the exact same job as the code on the left. I do not think the exception is needed. If it is, I think it would be better to catch the more specific issue coming from Exception and handle it accordingly with more detailed warnings. But I have not been able to think of a case, barring something really crazy hitting the server.

ruebot: > I believe the code on the right does the exact same job as the code on the left?

Are you talking about a diff? Otherwise I don't gather what you're referring to here with left and right.

greebie: Yeah - sorry.  In the view I was using the changed code was on the right and the old code was on the left.

greebie: More clearly, the `extract` method is incorporated into `apply` with no (known) differences in the way ExtractBoilerplateText works.

ianmilligan1: Tested - and it works.

Issue #194, opened 2018-04-06
More complete Twitter Ingestion


greebie:
Threaded from #192 

Going through the twitter ingestion script, it might be nice to include a few other things if possible from the api. Retweet_count for instance. Possibly things like "location" and "in_reply_to" depending on how far we want to go with that and ethical considerations.

Here's a short list from my perspective and maybe we can check the things we want to include, and then we can fix things all together:

- [x] in_reply_to_status_id
- [x] in_reply_to_user_id
- [x] source (mobile or web interface)
- [ ] entities/user_mentions
- [ ] entities/urls
- [ ] entities/hashtags (could replace the hashtag extractor we have in aut)
- [ ] ~retweeted (true / false)~
- [ ] place
- [x] retweet_count
- [ ] retweet_status { provides another tweet object so maybe not }
- [x] user/profile_image
- [x] user/description
- [x] user/time_zone
- [x] user/location
- [x] user/name
- [x] user/url

This is just a rough list of possible candidates. Not necessarily suggesting we should ingest them all or even one. Just think it's worth looking at while we fix TweetUtils.

ruebot:
`place` is an array. What do we want out of it?

Example:
```
$ zcat WomensMarch-20170123.json.gz | jq .place
{
  "full_name": "Holden, MA",
  "url": "https://api.twitter.com/1.1/geo/id/00173cd41f2b16d3.json",
  "country": "United States",
  "place_type": "city",
  "bounding_box": {
    "type": "Polygon",
    "coordinates": [
      [
        [
          -71.918466,
          42.312204
        ],
        [
          -71.918466,
          42.400332
        ],
        [
          -71.801795,
          42.400332
        ],
        [
          -71.801795,
          42.312204
        ]
      ]
    ]
  },
  "country_code": "US",
  "attributes": {},
  "id": "00173cd41f2b16d3",
  "name": "Holden"
}
```

ruebot:
All of the `entities` options are arrays too. So we'll have to map all those out as well.

```
  "entities": {
    "hashtags": [],
    "symbols": [],
    "user_mentions": [],
    "urls": []
  }
```

`user_mentions`

```
$ zcat WomensMarch-20170123.json.gz | head -n 1 | jq .entities.user_mentions
[
  {
    "indices": [
      0,
      15
    ],
    "id_str": "18395177",
    "screen_name": "solangeknowles",
    "name": "solange knowles",
    "id": 18395177
  },
  {
    "indices": [
[...]

PR #254, opened 2018-08-10
Add additional tweet fields to TweetUtils; partially address #194.

ruebot:
**GitHub issue(s)**: #194 

Adds the following addition fields to TweetUtils:

  - retweet_count
  - favorite_count
  - in_reply_to_status_id_str
  - in_reply_to_user_id_str
  - in_reply_to_screen_name
  - source
  - user.protected
  - user.profile_image_url
  - user.description
  - user.location
  - user.name
  - user.url
  - user.time_zone

# How should this be tested?

Same as outlined in https://github.com/archivesunleashed/aut/pull/252, but test additional fields.

Tests should take care of it all. But, definitely curious to see how it goes.

# Additional Notes:

Still need to add some addition fields outlined in #194. Entity fields will be interesting since they are all an array.

# Interested parties

@ianmilligan1 @greebie @lintool

### DIFF
--- a/src/main/scala/io/archivesunleashed/util/TweetUtils.scala
+++ b/src/main/scala/io/archivesunleashed/util/TweetUtils.scala
@@ -30,23 +30,49 @@ object TweetUtils {
   implicit class JsonTweet(tweet: JValue) {
     val user = "user"
     implicit lazy val formats = org.json4s.DefaultFormats
-    /** Get Twitter status id. */
+    /** Get tweet id. */
     def id(): String = try { (tweet \ "id_str").extract[String] } catch { case e: Exception => ""}
-    /** Get the date a status was created. */
+    /** Get the date a tweet was created. */
     def createdAt(): String = try { (tweet \ "created_at").extract[String] } catch { case e: Exception => ""}
-    /** Get the status text. */
+    /** Get the tweet text. */
     def text(): String = try { (tweet \ "text").extract[String] } catch { case e: Exception => ""}
-    /** Get the full_text. */
+    /** Get the tweet full_text. */
     def fullText(): String = try { (tweet \ "full_text").extract[String] } catch { case e: Exception => ""}
-    /** Get the language code (ISO 639-1). */
+    /** Get the tweet language code (ISO 639-1). */
     def lang: String = try { (tweet \ "lang").extract[String] } catch { case e: Exception => ""}
-    /** Get the username of the user who wrote the status. */
+    /** Get the username of the user who wrote the tweet. */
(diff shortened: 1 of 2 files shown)


src/test/scala/io/archivesunleashed/util/TweetUtilsTest.scala
```diff
@@ -73,5 +114,18 @@ class TweetUtilsTest extends FunSuite {
...
+      assert(tweet.inReplyToStatusId() == null)
+      assert(tweet.inReplyToUserId() == null)
+      assert(tweet.source() == null)
+      assert(tweet.isUserProtected() == false)
+      assert(tweet.profileImageUrl() == null)
+      assert(tweet.profileDescription() == null)
+      assert(tweet.profileLocation() == null)
+      assert(tweet.profileName() == null)
+      assert(tweet.profileUrl() == null)
+      assert(tweet.profileTimezone() == null)
```
greebie: Maybe use ".isEmpty" instead of == null, so scalastyle does not burp. Technically the return should be String instead of Null anyway.

ruebot: For all of them?

greebie: Just for the assertions. Although this is strictly a style issue, so not mandatory.  (e.g. `assert(tweet.profileLocation().isEmpty)`)

ruebot: It's out of scope for this PR, and should be a new ticket. Create a new ticket.

greebie: Hmm. That weird - these should have been detected and resolved in #249.  I will add the ticket and try to find out why scalastyle did not yell at me.

src/main/scala/io/archivesunleashed/util/TweetUtils.scala
```diff
@@ -30,23 +30,49 @@ object TweetUtils {
...
+    /** Get tweet id. */
     def id(): String = try { (tweet \ "id_str").extract[String] } catch { case e: Exception => ""}
-    /** Get the date a status was created. */
+    /** Get the date a tweet was created. */
     def createdAt(): String = try { (tweet \ "created_at").extract[String] } catch { case e: Exception => ""}
-    /** Get the status text. */
+    /** Get the tweet text. */
     def text(): String = try { (tweet \ "text").extract[String] } catch { case e: Exception => ""}
-    /** Get the full_text. */
+    /** Get the the tweet full_text. */
```
ianmilligan1: double "the the" here

Issue #253, opened 2018-08-10
Create UDF to combine basic text filtering features


greebie:
**Is your feature request related to a problem? Please describe.**

We have a number of features like RemoveHtml, RemoveHttpHeader and RemoveBoilerPlate that are fairly common features that people would like to use all at once in one call. We should provide a catchall do perform this common text operation.

**Describe the solution you'd like**
Provide a UDF that combines the main text filtering functions for analysis.

**Describe alternatives you've considered**
We could simply offer documentation that creates the function outside of aut.

**Additional context**
We discovered this after user feedback from the Slack channel.

ruebot:
Resolved with: https://github.com/archivesunleashed/aut/commit/84a4c094df4595ac88cfd50b5aaf2bb1f1d9cacc

PR #256, opened 2018-08-11
ExtractBoilerpipeText to remove headers as well. #253

greebie:
**GitHub issue(s)**:
#253

Alters ExtractBoilerpipeText to remove headers from text output as well.

# How should this be tested?

The test I created + Travis and Codecov should cover it.

# Additional Notes:

I attempted to provide a .keepHeader option, but to do so would require moving the .apply call from the object and that would require documentation changes. This seemed too much to accommodate a fairly marginal use-case.

This option can be provided by returning a class with .keepHeader and .text attributes. I do not think it's too necessary.

# Interested parties

@ianmilligan1

### DIFF
--- a/src/main/scala/io/archivesunleashed/matchbox/ExtractBoilerpipeText.scala
+++ b/src/main/scala/io/archivesunleashed/matchbox/ExtractBoilerpipeText.scala
@@ -19,10 +19,6 @@ package io.archivesunleashed.matchbox
 import de.l3s.boilerpipe.extractors.DefaultExtractor
 import java.io.IOException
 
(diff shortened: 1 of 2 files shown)


ianmilligan1:
I'm happy to merge if this looks good to you as well @ruebot.

ianmilligan1:
Nope, it's a straight change (just removes the http headers from the boilerplate removal, which is what we should have been doing all along) - tested w/ the documentation script and works well.

greebie:
Yes it does resolve.  I thought the solution was something else, but this seemed more straight-forward.

Issue #247, opened 2018-07-30
Method to perform finer-grained selection of ARCs and WARCs


lintool:
We currently only have one method to load ARCs or WARCs:

> `RecordLoader.loadArchives("/path/to/many/warcs/*.gz", sc)`

It'd be nice to have some fine grained control, e.g., I want all (W)ARCs starting with a prefix, _except_ for A, B, and C - this would help us debug large collections that maybe errors/corruption/etc.

borislin:
@lintool @ianmilligan1 

Based on our discussion on Slack, we would like to limit the size of individual records instead of archive files. A lot of Spark jobs fail because of records that are too big.

ianmilligan1:
Should this still be an open issue? I don't think we've been running into any ingestion issues lately, including on some very large collections.

ruebot:
I think the work done here is worth revisiting in the future, since I believe it got at the spirit of @lintool original post:

>It'd be nice to have some fine grained control, e.g., I want all (W)ARCs starting with a prefix, except for A, B, and C - this would help us debug large collections that maybe errors/corruption/etc.

But, I'll leave it up to him to close or keep it open.

ianmilligan1:
Sounds good, thanks @lintool @ruebot !

lintool:
Please leave open.

ruebot:
The spirit of this issue being identifying problematic W/ARCs will be solved with #533, so I'm going to close this once it is merged.

The additions to `loadArchives` [here](https://github.com/archivesunleashed/aut/pull/257/files#diff-7e582908ce37e25cdae381cebc539965c62f5a241cf1ea38fcafe9683b6ce44cR92-R97) still might be worth bringing in if there is a demand for it in the future.

PR #275, opened 2018-10-12
Refactor loadArchives() function to limit size of individual record

borislin:
**This PR refactors loadArchives() function to enable user to limit the size of individual record, preventing Spark jobs from failing due to large records.**

* * *

**GitHub issue(s)**:

* #247

This PR adds the functionality to the `loadArchives()` function to limit size of individual records to be processed, preventing Spark jobs from failing due to records that are too big.

# How should this be tested?

* `git fetch --all`
* `git checkout limit-record-size`
* `mvn clean install`
* Create an output directory with sub-directories:
`mkdir -p path/to/where/ever/you/can/write/output/all-domains path/to/where/ever/you/can/write/spark-jobs`
* Adapt the script below:
```
import io.archivesunleashed._
sc.setLogLevel("INFO")

//val input = "/tuna1/scratch/aut-issue-271/warcs/*.gz"
val input = "/tuna1/scratch/aut-issue-271/warcs/ARCHIVEIT-499-MONTHLY-JOB311846-20170703003948338-00020.warc.gz"

val output1 = "/tuna1/scratch/aut-issue-271/derivatives/all-domains"

val maxRecordSize: Option[Long] = Some(174552L)

RecordLoader.loadArchives(input, sc, maxRecordSize).keepValidPages().map(r => r.getContentBytes.length).sortBy(r => r, ascending = false).saveAsTextFile(output1)

sys.exit
```

* Run the following command with adapted paths with Apache Spark 2.1.3:
`/home/b25lin/spark-2.1.3-bin-hadoop2.7/bin/spark-shell --master local[10] --driver-memory 30G --conf spark.network.timeout=100000000 --conf spark.executor.heartbeatInterval=6000s --conf spark.driver.maxResultSize=10G --jars "/tuna1/scratch/borislin/aut/target/aut-0.17.1-SNAPSHOT-fatjar.jar" -i /tuna1/scratch/aut-issue-271/spark_jobs/499.scala | tee /tuna1/scratch/aut-issue-271/spark_jobs/499.scala.log`

* Change the value of the `maxRecordSize` variable in the above script to a different value or remove it. See different results being produced. Only records that have sizes that are smaller or equal to the user-provided `maxRecordSize` will be filtered out.

# Additional Notes:

* This PR doesn't affect our current use and documentation for the `loadArchives()` function since `maxRecordSize` is an optional parameter. But we should update our documentation to reflect that now user can use this `maxRecordSize` variable to limit record size.

# Results

My results are in `/tuna1/scratch/aut-issue-271/derivatives` and log is in `/tuna1/scratch/aut-issue-271/spark_jobs/499.scala.log`

# Interested parties

@lintool @ianmilligan1 @ruebot

### DIFF
--- a/src/main/scala/io/archivesunleashed/package.scala
+++ b/src/main/scala/io/archivesunleashed/package.scala
@@ -23,6 +23,7 @@ import io.archivesunleashed.matchbox.{ComputeMD5, DetectLanguage, ExtractDate, E
 import io.archivesunleashed.matchbox.ImageDetails
 import io.archivesunleashed.matchbox.ExtractDate.DateComponent
 import org.apache.hadoop.fs.{FileSystem, Path}
+import org.apache.log4j.Logger
 // scalastyle:off underscore.import
 import io.archivesunleashed.matchbox.ExtractDate.DateComponent._
 import org.apache.spark.sql._
@@ -42,6 +43,9 @@ import scala.util.matching.Regex
 package object archivesunleashed {
   /** Loads records from either WARCs, ARCs or Twitter API data (JSON). */
   object RecordLoader {
+
+    val log: Logger = Logger.getLogger(getClass.getName)
+
     /** Gets all non-empty archive files.
       *
       * @param dir the path to the directory containing archive files
@@ -51,6 +55,7 @@ package object archivesunleashed {
     def getFiles(dir: Path, fs: FileSystem): String = {
       val statuses = fs.globStatus(dir)
       val files = statuses.filter(f => fs.getContentSummary(f.getPath).getLength > 0).map(f => f.getPath)
+      log.info("Number of archive files: " + files.length)
       files.mkString(",")
     }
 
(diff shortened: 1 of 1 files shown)


lintool:
@ruebot - @ianmilligan1 and I discussed this as a general strategy for dealing with edge-case WARCs and ARCs. Previously, we've dealt with the issue by increasing timeouts, which is janky. This seems like a cleaner solution.

@borislin Are you sure something like this works? IIRC, by the time we parse the header, we've already committed to parsing the entire record, which defeats the whole point of filtering by size. I believe this size check needs to be pushed further down into the reader.

The way to test this is to go back and examine those IA ARCS we failed on - stored on `tuna`.

borislin:
@ianmilligan1 I'm running aut with this PR to see whether we still have those heartbeat issues in IA ARCs. I think I may need to update this PR after that. Will keep this PR updated once I have more results...it's very slow to validate all those ARCs.

ianmilligan1:
OK - please keep us posted on that front. I'll make sure to re-test if the PR changes.

borislin:
Update:

I've successfully run AUT (based on current `master` branch) on all IA ARCs and WARCs as well as `/tuna1/scratch/archiveit/739/4532` to obtain MIME, and have not observed any more heartbeat timeout issues. I think we do not need to refactor the `loadArchives()` function in this PR for now.

All MIME results, scripts and log files are at `/tuna1/scratch/ia-issue-1-arc-{arc_num}/` (0 <= `{arc_num}` <= 5)  and `/tuna1/scratch/ia-issue-1-warc-2/`.

lintool:
Per my comment above dated Oct 12 - I don't think this is addressing the issue at the right level in the stack. Furthermore, general timeout issues related to WARC/ARC processing seems to have been fixed. Closing this issue and resolving as "won't fix".

Issue #243, opened 2018-07-20
Replace hashing of unique ids with .zipWithUniqueId()


greebie:
**Describe the Enhancement**

AUT uses hash values to create unique ids, which can leave us duplicates of the same url in a network graph when hashes collide.

**To Reproduce**
Steps to reproduce the behavior (e.g.):
Run a Domain Graph Extractor with a large number of network nodes (websites).
Run in Gephi.
Discover duplicate websites in graph.

**Expected behavior**
All network nodes should be unique.

**Screenshots**
N/A

**Additional context**
The .zipWithIndex() feature in Apache Spark would be a better approach. http://spark.apache.org/docs/latest/api/python/pyspark.html#pyspark.RDD.zipWithIndex

.zipWithUniqueId() does not call another context so it could be faster.

See also #228

greebie:
I found a solution, which still needs testing, but current time trials using this code:

```
def timed(f: => Unit) = {
  val start = System.currentTimeMillis()
  f
  val end = System.currentTimeMillis()
  println("Elapsed Time: " + (end - start))
}

timed {
import io.archivesunleashed._
import io.archivesunleashed.app._
import io.archivesunleashed.matchbox._

val links = RecordLoader.loadArchives("/Users/ryandeschamps/warcs/", sc)
  .keepValidPages()
  .map(r => (r.getCrawlDate, ExtractLinks(r.getUrl, r.getContentString)))
  .flatMap(r => r._2.map(f => (r._1, ExtractDomain(f._1).replaceAll("^\\s*www\\.", ""), ExtractDomain(f._2).replaceAll("^\\s*www\\.", ""))))
  .filter(r => r._2 != "" && r._3 != "")
  .countItems()
  .filter(r => r._2 > 5)

WriteGEXF(links, "new-gephi3.gexf") }
```

ianmilligan1:
Thanks for this @greebie! This new approach seems better, but I'm just weighing the cost/benefit of longer processing time vs. avoiding all hash collisions. 

Right now, when we have collisions, Gephi automatically fixes those is that correct?

greebie:
Not really. Gephi merges items that have the collisions, so with large collections, it could produce a misrepresentation of the graph. However, I have not come across it in this example.

ianmilligan1:
OK great, thanks for the explanations and documentation here. My sense is that this is a better approach, but I'm not 100% due to the time trade-off.

Thoughts @ruebot @lintool ?

[...]

PR #289, opened 2018-11-07
Change Id generation for graphs from using hashes for urls to using .zipWithUniqueIds()

greebie:
**GitHub issue(s)**:

#243

This PR adds WriteGraph which generates GEXF and/or Graphml files.
There are also some additional utilities such as an id lookup.

This object generates proper unique ids. This differs from the current method in WriteGraphML and WriteGEXF that simply creates ids using an MD5 hash of the url.  This method is better because the latter method could produce incorrect graphs in the case where hashes collide (eg. with very large graphs). 

Timing tests with a medium-sized graph shows that it increases the processing time by 10-15%. However, it could reduce the size of network graph derivatives by an unknown margin (because numeric ids are smaller than hashes).

Example:
Old way produces:
```
<node id="405d19a958ba43d88e9edd7a77338aa3" label="laws.justice.gc.ca" />
<node id="69bde2cbb357119a50a950fca99a8341" label="english.uvic.ca" />
<node id="8687e4dc11548a5917504975feb7c649" label="oipc.bc.ca" />
<node id="30192bae759dafccc58bccc268c2b411" label="accuweather.com" />
```
New way:
```
<node id="0" label="laws.justice.gc.ca" />
<node id="38" label="english.uvic.ca" />
<node id="76" label="oipc.bc.ca" />
<node id="114" label="accuweather.com" />
```
# How should this be tested?

- Travis should pass.

```
def timed(f: => Unit) = {
  val start = System.currentTimeMillis()
  f
  val end = System.currentTimeMillis()
  println("Elapsed Time: " + (end - start))
}

timed {
import io.archivesunleashed._

val links = RecordLoader.loadArchives("/Users/ryandeschamps/warcs/", sc)
  .keepValidPages()
  .map(r => (r.getCrawlDate, ExtractLinks(r.getUrl, r.getContentString)))
  .flatMap(r => r._2.map(f => (r._1, ExtractDomain(f._1).replaceAll("^\\s*www\\.", ""), ExtractDomain(f._2).replaceAll("^\\s*www\\.", ""))))
  .filter(r => r._2 != "" && r._3 != "")
  .countItems()
  .filter(r => r._2 > 5)

WriteGraph(links, "new-gephi3.gexf") }
```

Should produce a Gexf that opens in Gephi.

Change last line to ```WriteGraph.asGraphml(links, "new-gephi3.gexf")``` to get graphml.

I have not tested it in GraphPass yet, but there is no reason why it should not work as directed.

# Additional Notes:

- `WriteGraphml`would replace WriteGEXF & WriteGraphml which can be deprecated.
- `CommandLineApp` would also have to be changed before WriteGEXF etc. are removed.
- I think aut had a previous `WriteGraph` udf which was deprecated and removed.
- This applies only to the RDD graph functions.  DF functions are not changed. (I need to review the way DFs do unique ids)
- I added a node id lookup tool during the process of developing WriteGraph. It might be useful for the toolkit, but it can also be removed.  

# Interested parties

@lintool @ruebot @ianmilligan1 

### DIFF
--- a/src/main/scala/io/archivesunleashed/app/WriteGraph.scala
+++ b/src/main/scala/io/archivesunleashed/app/WriteGraph.scala
@@ -31,7 +31,7 @@ object WriteGraph {
   val nodeStart = "<node id=\""
   val edgeStart = "<edge source=\""
   val targetChunk = "\" target=\""
-  val directedChunk = " type=\"directed\">\n"
+  val directedChunk = "\" type=\"directed\">\n"
   val endAttribute = "\" />\n"
   val labelStart = "\" label=\""
   val xmlHeader = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
@@ -139,7 +139,7 @@ object WriteGraph {
         edgeStart  +
         sid + targetChunk +
         did + "\" weight=\"" + weight +
-        "\" type=\"directed\">\n" +
+        directedChunk +
         "<attvalues>\n" +
         "<attvalue for=\"0\" value=\"" + date + endAttribute +
         "</attvalues>\n" +
--- a/src/main/scala/io/archivesunleashed/app/WriteGraph.scala
+++ b/src/main/scala/io/archivesunleashed/app/WriteGraph.scala
@@ -59,7 +59,7 @@ object WriteGraph {
   }
 
   def apply(ds: Array[Row], gexfPath: String): Boolean = {
-    if (!gexfPath.isEmpty())  { asGexf(ds, gexfPath) } else { false }
+    if (!gexfPath.isEmpty())  { asGexf (ds, gexfPath) } else { false }
   }
 
   /** Produces a zipped RDD with ids and labels from an network-based RDD, rdd.
@@ -120,7 +120,7 @@ object WriteGraph {
    * @return true on success.
    */
   def asGexf (rdd: RDD[((String, String, String), Int)], gexfPath: String): Boolean = {
-    if (!gexfPath.isEmpty()) {
+    if (gexfPath.isEmpty()) {
       false
     } else {
       val nodes = nodesWithIds(rdd)
--- a/src/main/scala/io/archivesunleashed/app/WriteGraph.scala
+++ b/src/main/scala/io/archivesunleashed/app/WriteGraph.scala
@@ -55,20 +55,11 @@ object WriteGraph {
    * @return Unit().
    */
   def apply(rdd: RDD[((String, String, String), Int)], gexfPath: String): Boolean = {
-    if (gexfPath.isEmpty()) {
-      false
-    } else
-    {
-      asGexf (rdd, gexfPath)
-    }
+    if (!gexfPath.isEmpty()) { asGexf (rdd, gexfPath) } else { false }
   }
 
   def apply(ds: Array[Row], gexfPath: String): Boolean = {
-    if (gexfPath.isEmpty())  {
-      false
-    } else {
-      asGexf (ds, gexfPath)
-    }
+    if (!gexfPath.isEmpty())  { asGexf(ds, gexfPath) } else { false }
   }
 
   /** Produces a zipped RDD with ids and labels from an network-based RDD, rdd.
@@ -129,10 +120,9 @@ object WriteGraph {
    * @return true on success.
    */
   def asGexf (rdd: RDD[((String, String, String), Int)], gexfPath: String): Boolean = {
-    if (gexfPath.isEmpty()) {
+    if (!gexfPath.isEmpty()) {
       false
-    } else
-    {
+    } else {
       val nodes = nodesWithIds(rdd)
       val outFile = Files.newBufferedWriter(Paths.get(gexfPath), StandardCharsets.UTF_8)
       val edges = edgeNodes(rdd).map({ case (date, sid, did, weight) =>
@@ -171,11 +161,10 @@ object WriteGraph {
   def asGexf (data: Array[Row], gexfPath: String): Boolean = {
     if (gexfPath.isEmpty()) {
       false
-    } else
-    {
+    } else {
       val outFile = Files.newBufferedWriter(Paths.get(gexfPath), StandardCharsets.UTF_8)
       val vertices = scala.collection.mutable.Set[String]()
-      data foreach { d =>
+      data.foreach { d =>
         vertices.add(d.get(1).asInstanceOf[String])
         vertices.add(d.get(2).asInstanceOf[String])
       }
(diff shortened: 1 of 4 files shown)


greebie:
Just a few additional notes here.

ExtractGraphX still uses hash and not .zipWithUniqueId().  It's possible that this graph generation approach is slightly faster due to optimization (when not calculating PageRank etc.).  I'm going to explore more now that I'm coming closer to the end of term with teaching.

ianmilligan1:
>I'm going to explore more now that I'm coming closer to the end of term with teaching.

Would that be in a separate PR or would it potentially affect this one?

greebie:
I think it should be a separate PR, possibly referencing a different issue. I've been thinking about ExtractGraphX as a way to reduce the problems with GraphPass, but it's possible we could see some small efficiency gains just for regular aut production.  I'd like to test that out.

ianmilligan1:
OK sounds good @greebie. I'll let @lintool and @ruebot review this. From a user perspective, works for me.

src/main/scala/io/archivesunleashed/app/WriteGraph.scala
```diff
@@ -0,0 +1,239 @@
...
+    "  http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd\"\n>"
+
+  /** Writes graph nodes and edges to file.
+   *
+   * @param rdd RDD of elements in format ((datestring, source, target), count)
+   * @param gexfPath output file
+   * @return Unit().
+   */
+  def apply(rdd: RDD[((String, String, String), Int)], gexfPath: String): Boolean = {
+    if (gexfPath.isEmpty()) {
```
lintool: `if (!gexfPath.isEmpty()) asGexf(rdd, gexfPath) else false`

src/main/scala/io/archivesunleashed/app/WriteGraph.scala
```diff
@@ -0,0 +1,239 @@
...
+   *
+   * @param rdd a RDD of elements in format ((datestring, source, target), count)
+   * @param gexfPath output file
+   * @return true on success.
+   */
+  def asGexf (rdd: RDD[((String, String, String), Int)], gexfPath: String): Boolean = {
+    if (gexfPath.isEmpty()) {
+      false
+    } else
+    {
```
lintool: braces on previous line

src/main/scala/io/archivesunleashed/app/WriteGraph.scala
```diff
@@ -0,0 +1,239 @@
...
+    *
+    * @param data a Dataset[Row] of elements in format (datestring, source, target, count)
+    * @param gexfPath output file
+    * @return true on success.
+    */
+  def asGexf (data: Array[Row], gexfPath: String): Boolean = {
+    if (gexfPath.isEmpty()) {
+      false
+    } else
+    {
```
lintool: braces with previous line.

src/main/scala/io/archivesunleashed/app/WriteGraph.scala
```diff
@@ -0,0 +1,239 @@
...
+      val outFile = Files.newBufferedWriter(Paths.get(gexfPath), StandardCharsets.UTF_8)
+      val vertices = scala.collection.mutable.Set[String]()
+      data foreach { d =>
```
lintool: lets use dot notations to be consistent with other collection operations? same below.

src/main/scala/io/archivesunleashed/app/WriteGraph.scala
```diff
@@ -0,0 +1,239 @@
...
+      outFile.write("</edges>\n</graph>\n</gexf>")
+      outFile.close()
+      true
+    }
+  }
+
+  def asGraphml (rdd: RDD[((String, String, String), Int)], graphmlPath: String): Boolean = {
+    if (graphmlPath.isEmpty()) {
+      false
+    } else
```
lintool: braces with previous line.

src/test/scala/io/archivesunleashed/app/WriteGraphTest.scala
```diff
@@ -0,0 +1,146 @@
...
+                         (("Date3", "Source3", "Destination3"), 100))
+  private val testFile = "temporaryTestFile.txt"
+
+  before {
+    val conf = new SparkConf()
+      .setMaster(master)
+      .setAppName(appName)
+      conf.set("spark.driver.allowMultipleContexts", "true");
+      sc = new SparkContext(conf)
+    }
```
lintool: wrong indentation level

ianmilligan1:
I got a failure running this on rho:

```scala
import io.archivesunleashed._

val links = RecordLoader.loadArchives("/mnt/vol1/data_sets/geocities/warcs/indexed/GEOCITIES-20091029185858-00184-crawling08.us.archive.org.warc.gz", sc)
  .keepValidPages()
  .map(r => (r.getCrawlDate, ExtractLinks(r.getUrl, r.getContentString)))
  .flatMap(r => r._2.map(f => (r._1, ExtractDomain(f._1).replaceAll("^\\s*www\\.", ""), ExtractDomain(f._2).replaceAll("^\\s*www\\.", ""))))
  .filter(r => r._2 != "" && r._3 != "")
  .countItems()
  .filter(r => r._2 > 5)

WriteGraph(links, "/home/i2millig/results/new-gephi.gexf")
```

Did syntax change in the revisions or is there an issue? [Error log is here](https://gist.github.com/ianmilligan1/e5a1efe1b55f27e67397bfaee0629ea6) but the tl;dr is:

```
scala.MatchError: ("http,((1042,110mb.com,20091029,9),None)) (of class scala.Tuple2)
        at io.archivesunleashed.app.WriteGraph$$anonfun$edgeNodes$5.apply(WriteGraph.scala:112)
        at io.archivesunleashed.app.WriteGraph$$anonfun$edgeNodes$5.apply(WriteGraph.scala:112)
```

greebie:
It looks like the problem is that on converting the edge tuples to ids, I did not escape the xml first.  That's why the wonky `"http` caused problems.  Can I borrow a copy of the geocities file so I can try to duplicate the problem and then test it for later versions?

ianmilligan1:
Am sending the WARC to you via Slack, @greebie.

As we've now had a number of fails on this, could you also do a close line-by-line read of your new `WriteGraph` function vs `WriteGEXF` and `WriteGraphML` - the latter two are quite robust, so we'll want to make sure that the new one is too. i.e. a checklist of what `WriteGEXF` does and then making sure those are mapped over.

greebie:
The bigger problem is that WriteGEXFTest and WriteGraphmlTest did not test for unescaped xml in the result.  It tested the xml escaping function, but not that it was applied in the result. All these errors should be caught at compile time.

greebie:
Ran the same code on the new graph and it worked (after a java heap space error, but that's not related to this).

I also loaded the graph into gephi and it worked with no errors.

ianmilligan1:
OK worked here too. Can you confirm that there's nothing missing in this `WriteGraph` function that is present in the other two?

greebie:
I reviewed the code and it works as expected, based on my knowledge of the circumstances.  I have one more push, however, to include some additional tests to cover potential future problems.

ianmilligan1:
Ok ready for me to review again on your end @greebie?

Issue #198, opened 2018-04-20
Function to get the status response code and headers of a warc response?


dportabella:
is there a function to get the status response code and headers of a warc response?

such as...
```
RecordLoader.loadArchives(warcFile, sc)
.filter(_.warcResponse.statusCode = 200)
.filter(_.warcResponse.headers.get("Server") == "Apache/2.4.6")
```

dportabella:
I am using this function at the moment:

```
import java.io.ByteArrayInputStream
import io.archivesunleashed.spark.archive.io.ArchiveRecord
import org.apache.commons.httpclient.{Header, HttpParser, StatusLine}
import org.apache.commons.io.IOUtils

case class Response(archiveRecord: ArchiveRecord, statusLine: StatusLine, headers: List[Header], content: Array[Byte])

object WarcUtils {
  def parseResponse(r: ArchiveRecord): Response = {
    val response = new ByteArrayInputStream(r.getContentBytes)
    val line = HttpParser.readRawLine(response)
    val statusLine = new StatusLine(new String(line))
    val headers = HttpParser.parseHeaders(response, "US-ASCII").toList
    val responseContent: Array[Byte] = IOUtils.toByteArray(response)
    Response(r, statusLine, headers, responseContent)
  }
}

RecordLoader.loadArchives(warcFile, sc)
  .filter(FilterArchive.isHTML)
  .flatMap(r => Try(WarcUtils.parseResponse(r)).toOption)
  .filter(_.statusLine.getStatusCode = 200)
  .filter(_.headers.collectFirst {case h if h.getName == "Server" => h.getValue }.contains("Apache/2.4.6"))
```

dportabella:
Although there is an error parsing the following archive, generate by
`$ wget --warc-file=test https://www.linkedin.com/`

do you know if this `1000` (just before the html content) is part of the response header? it looks quite strange to me. my `parseResponse` function detects this `1000` as part of the html content.

any idea about this?

```
WARC/1.0
WARC-Type: response
WARC-Record-ID: <urn:uuid:5D495122-1DD9-472B-B619-A7EDADF03070>
WARC-Warcinfo-ID: <urn:uuid:65D43BDC-AC23-4515-A951-6DE201680871>
WARC-Concurrent-To: <urn:uuid:CB15E601-0128-4D95-9436-BDF271F23CBF>
WARC-Target-URI: <https://www.linkedin.com/>
WARC-Date: 2018-04-20T07:57:54Z
[...]

Issue #164, opened 2018-01-22
feature request: ArchiveRecord.archiveFile


dportabella:
I am querying CommonCrawl archive, which is divided into hundreds of warc.gz files. I use RecordLoader.loadArchives to read all the warc files at once. Sometimes the log contains an Exception when processing a page, and I'd need to find out from which of the individual warc.gz files comes from (so that I can re-run the program in that file only).

Would it be possible for `ArchiveRecord` class to have also a field with the input archive name? (with that, I could catch exceptions and show not only the url but also the input archive file).

ianmilligan1:
Thanks as always (we're getting a bit of a back log of issues as we're stretched in many different directions, so don't take lack of work as lack of interest!). 

Just so I am clear, the idea would be that you'd have a command like:

```
  .map(r => (r.getArchiveFile, r.getCrawlDate, r.getDomain, r.getUrl, RemoveHTML(r.getContentString)))
```

Where `getArchiveFile` would then produce the WARC file string?

dportabella:
Hi, i am not sure I understand your question. here it is an example:

```
$ find /CC-MAIN-2016-36
/CC-MAIN-2016-36/segments/1471982931818.60/warc/CC-MAIN-20160823200851-00286-ip-10-153-172-175.ec2.internal.warc.gz
/CC-MAIN-2016-36/segments/1471982931818.60/warc/CC-MAIN-20160823200851-00140-ip-10-153-172-175.ec2.internal.warc.gz
...
/CC-MAIN-2016-36/segments/1471982292675.36/warc/CC-MAIN-20160823195812-00089-ip-10-153-172-175.ec2.internal.warc.gz
/CC-MAIN-2016-36/segments/1471982292675.36/warc/CC-MAIN-20160823195812-00190-ip-10-153-172-175.ec2.internal.warc.gz
...

RecordLoader.loadArchives("/CC-MAIN-2016-36", sc)
.foreach(r => println(r.getArchiveFile + "\t" + r.getUrl)


/CC-MAIN-2016-36/segments/1471982931818.60/warc/CC-MAIN-20160823200851-00286-ip-10-153-172-175.ec2.internal.warc.gz   http://example.com/p1.html
/CC-MAIN-2016-36/segments/1471982931818.60/warc/CC-MAIN-20160823200851-00286-ip-10-153-172-175.ec2.internal.warc.gz   http://example.com/p2.html
[...]

Issue #260, opened 2018-08-14
Improve ArchiveRecord.scala test coverage


ruebot:
[`ArchiveRecord.scala`](https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/ArchiveRecord.scala) [test coverage](https://codecov.io/gh/archivesunleashed/aut/src/master/src/main/java/io/archivesunleashed/data/ArchiveRecordWritable.java) can be improved.

PR #292, opened 2018-11-22
Add .getHttpStatus and .getFilename to ArchiveRecordImpl class #198 & #164

greebie:
**GitHub issue(s)**:

#198, #164

This PR includes three changes to the ArchiveRecord class:
1. The Status Code Header for the crawl (via .getHttpStatus)
2. The read-streamed filename (via .getFilename)
3. Explicit testing of the ArchiveRecord class (this will likely not change coverage by much if at all).

`.getHttpStatus` could be useful for studies that would like to compare status calls on crawls.  

Example:

```scala
  import io.archivesunleashed._
  import io.archivesunleashed.util._

  val links = RecordLoader.loadArchives("/Users/ryandeschamps/warcs/*gz", sc)
    .keepValidPages()
    .keepContent(Set("apple".r))
    .map(r => (r.getHttpStatus, (ExtractLinks(r.getUrl, r.getContentString))))
    .flatMap(r => r._2.map(f => (r._1, ExtractDomain(f._1).replaceAll("^\\s*www\\.", ""), 
      ExtractDomain(f._2).replaceAll("^\\s*www\\.", ""))))
    .filter(r => r._2 != "" && r._3 != "")
    .countItems()
    .filter(r => r._2 > 5).take(10)
```
Should produce something like 

```
Array(((200,nanaimodailynews.com,nanaimodailynews.com),445785), ((200,nanaimodailynews.com,blackpress.ca),188676), ((200,nanaimodailynews.com,bclocalnews.com),111400), ... )
```

It might be more interesting to test this script without `.keepValidPages()` since most valid pages would likely return a `200`.  Where an ArchiveRecord fails to get a status code (via null or empty string) the record will return `000`.  This default value can be changed.

`.getFilename` returns the fullpath (which may be a url) of the Warc that was consumed.

```scala
  import io.archivesunleashed._
  import io.archivesunleashed.util._
  import org.apache.commons.io.FilenameUtils

  val links = RecordLoader.loadArchives("/Users/ryandeschamps/warcs/*gz", sc)
        .keepValidPages()
        .keepContent(Set("apple".r))
        .map(r => (r.getFilename, (ExtractLinks(r.getUrl, r.getContentString))))
        .flatMap(r => r._2.map(f => (r._1, ExtractDomain(f._1).replaceAll("^\\s*www\\.", ""), 
          ExtractDomain(f._2).replaceAll("^\\s*www\\.", ""))))
        .filter(r => r._2 != "" && r._3 != "")
        .countItems()
        .filter(r => r._2 > 5).take(10)

```

should produce something like 

```
links: Array[((String, String, String), Int)] = Array(((file:/Users/ryandeschamps/warcs/ARCHIVEIT-4656-CRAWL_SELECTED_SEEDS-JOB193391-20160127222913427-00000.warc.gz,nanaimodailynews.com,nanaimodailynews.com),439503), 
((file:/Users/ryandeschamps/warcs/ARCHIVEIT-4656-CRAWL_SELECTED_SEEDS-JOB193391-20160127222913427-00000.warc.gz,nanaimodailynews.com,blackpress.ca),186028), 
((file:/Users/ryandeschamps/warcs/ARCHIVEIT-4656-CRAWL_SELECTED_SEEDS-JOB193391-20160127222913427-00000.warc.gz,nanaimodailynews.com,bclocalnews.com),106107), 
((file:/Users/ryandeschamps/warcs/ARCHIVEIT-4656-CRAWL_SELECTED_SEEDS-JOB193391-20160127222913427-00000.warc.gz,nanaimodailynews.com,drivewaycanada.ca),53040),
...
```

To get just the filename, you could use `FilenameUtils.getName(x.getFilename)` (I have included the FilenameUtils import in the code above).  I have not looked into whether FilenameUtils will get the filename from a url, but that is why I stuck with the fullpath. Also, I am open to feedback on whether .getFilename is the right call for this.  IIPC calls it `.getReaderIdentifier()`.

I can think of a number of use cases for this, but it would be great as a way to detect problems with warcs (e.g. with wonky data).  It might also be helpful for error responses down the road.

# How should this be tested?

The above scripts should have expected outputs.
Both functions should work for both arc and warc formats.
Travis should pass.

# Additional Notes:

I used some of @dportabella 's ideas to produce the code here, in particular for getting the HttpStatus from a Warc file. For Arc and the Warc filename, I tried to use the WARCRecord and ARCRecord classes instead.

I did not include the full header responses in this PR. It seems like the ArcRecord and WarcRecord responses are quite different and I haven't been able to produce an effective testing mechanism.

# Interested parties

@ianmilligan1 @ruebot 

### DIFF
--- a/src/main/scala/io/archivesunleashed/ArchiveRecord.scala
+++ b/src/main/scala/io/archivesunleashed/ArchiveRecord.scala
@@ -31,8 +31,8 @@ import org.apache.commons.httpclient.{Header, HttpParser, StatusLine}
 
 /** Trait for a record in a web archive. */
 trait ArchiveRecord extends Serializable {
-  /** Returns the filename containing the Archive Records */
-  def getFilename: String
+  /** Returns the full path or url containing the Archive Records. */
+  def getResourcename: String
 
   /** Returns the crawl date. */
   def getCrawlDate: String
@@ -83,7 +83,7 @@ class ArchiveRecordImpl(r: SerializableWritable[ArchiveRecordWritable]) extends 
   }
   val ISO8601 = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssX")
 
-  val getFilename: String = {
+  val getResourcename: String = {
     if (r.t.getFormat == ArchiveRecordWritable.ArchiveFormat.ARC){
       arcRecord.getMetaData.getReaderIdentifier()
     } else {
--- a/src/main/scala/io/archivesunleashed/ArchiveRecord.scala
+++ b/src/main/scala/io/archivesunleashed/ArchiveRecord.scala
@@ -32,7 +32,7 @@ import org.apache.commons.httpclient.{Header, HttpParser, StatusLine}
 /** Trait for a record in a web archive. */
 trait ArchiveRecord extends Serializable {
   /** Returns the filename containing the Archive Records */
-  def getFileName: String
+  def getFilename: String
 
   /** Returns the crawl date. */
   def getCrawlDate: String
@@ -83,7 +83,7 @@ class ArchiveRecordImpl(r: SerializableWritable[ArchiveRecordWritable]) extends 
   }
   val ISO8601 = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssX")
 
-  val getFileName: String = {
+  val getFilename: String = {
     if (r.t.getFormat == ArchiveRecordWritable.ArchiveFormat.ARC){
       arcRecord.getMetaData.getReaderIdentifier()
     } else {
--- a/src/main/scala/io/archivesunleashed/ArchiveRecord.scala
+++ b/src/main/scala/io/archivesunleashed/ArchiveRecord.scala
@@ -18,6 +18,7 @@
 package io.archivesunleashed
 
 import java.text.SimpleDateFormat
+import java.io.ByteArrayInputStream
 
 import io.archivesunleashed.data.{ArcRecordUtils, WarcRecordUtils, ArchiveRecordWritable}
 import io.archivesunleashed.matchbox.{ExtractDate, ExtractDomain, RemoveHttpHeader}
@@ -25,6 +26,8 @@ import org.apache.spark.SerializableWritable
 import org.archive.io.arc.ARCRecord
 import org.archive.io.warc.WARCRecord
 import org.archive.util.ArchiveUtils
+import scala.util.Try
+import org.apache.commons.httpclient.{Header, HttpParser, StatusLine}
 
 /** Trait for a record in a web archive. */
 trait ArchiveRecord extends Serializable {
@@ -51,6 +54,10 @@ trait ArchiveRecord extends Serializable {
 
   /** Returns a raw array of bytes for an image. */
   def getImageBytes: Array[Byte]
+
+  /** Returns the http status of the crawl. */
+  def getHttpStatus: String
+
 }
 
 /** Default implementation of a record in a web archive.
@@ -64,6 +71,7 @@ class ArchiveRecordImpl(r: SerializableWritable[ArchiveRecordWritable]) extends 
   var arcRecord: ARCRecord = null
   var warcRecord: WARCRecord = null
   // scalastyle:on null
+  var headerResponseFormat: String = "US-ASCII"
 
   if (r.t.getFormat == ArchiveRecordWritable.ArchiveFormat.ARC) {
     arcRecord = r.t.getRecord.asInstanceOf[ARCRecord]
@@ -107,9 +115,10 @@ class ArchiveRecordImpl(r: SerializableWritable[ArchiveRecordWritable]) extends 
 
   val getMimeType: String = {
     if (r.t.getFormat == ArchiveRecordWritable.ArchiveFormat.ARC) {
-      arcRecord.getMetaData.getMimetype
+      Option(arcRecord.getMetaData.getMimetype).getOrElse("unknown")
     } else {
-      WarcRecordUtils.getWarcResponseMimeType(getContentBytes)
+      Option(WarcRecordUtils.getWarcResponseMimeType(getContentBytes))
+        .getOrElse("unknown")
     }
   }
 
@@ -118,6 +127,19 @@ class ArchiveRecordImpl(r: SerializableWritable[ArchiveRecordWritable]) extends 
       arcRecord.getMetaData.getUrl
     } else {
       warcRecord.getHeader.getUrl
+    }
+  }
+
+  val getHttpStatus: String = {
+    if (r.t.getFormat == ArchiveRecordWritable.ArchiveFormat.ARC) {
+      Option(arcRecord.getMetaData.getStatusCode).getOrElse("000")
+    } else {
+      Try(new StatusLine(new String(HttpParser.readRawLine
+        (new ByteArrayInputStream(getContentBytes))))
+        .getStatusCode).toOption match {
+          case Some(x) => x.toString
+          case None => "000"
+        }
     }
   }
 
--- a/src/main/scala/io/archivesunleashed/ArchiveRecord.scala
+++ b/src/main/scala/io/archivesunleashed/ArchiveRecord.scala
@@ -32,7 +32,7 @@ import org.apache.commons.httpclient.{Header, HttpParser, StatusLine}
 /** Trait for a record in a web archive. */
 trait ArchiveRecord extends Serializable {
   /** Returns the full path or url containing the Archive Records. */
-  def getResourcename: String
+  def getArchiveFilename: String
 
   /** Returns the crawl date. */
   def getCrawlDate: String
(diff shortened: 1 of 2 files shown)


ianmilligan1:
Thanks @greebie – I can test. Is this PR complete or is there anything else you're hoping to add to it before we start the review process?

greebie:
Thanks.  I'm pretty sure I learned from the last PR and did a very good job testing. *fingers crossed*  I think it's ready to go.

ianmilligan1:
OK sg. I'll take first round testing, realistically probably tomorrow morning.

src/test/scala/io/archivesunleashed/ArchiveRecordTest.scala
```diff
@@ -37,17 +38,76 @@ class ArchiveRecordTest extends FunSuite with BeforeAndAfter {
       .setAppName(appName)
     conf.set("spark.driver.allowMultipleContexts", "true");
     sc = new SparkContext(conf)
+
```
ruebot: Remove blank line.

src/main/scala/io/archivesunleashed/ArchiveRecord.scala
```diff
@@ -18,16 +18,22 @@
...
 import org.archive.io.arc.ARCRecord
 import org.archive.io.warc.WARCRecord
 import org.archive.util.ArchiveUtils
+import scala.util.Try
+import org.apache.commons.httpclient.{Header, HttpParser, StatusLine}
 
 /** Trait for a record in a web archive. */
 trait ArchiveRecord extends Serializable {
+  /** Returns the filename containing the Archive Records */
+  def getFilename: String
```
ruebot: Let's use `Resourcename` instead of `Filename``.

src/main/scala/io/archivesunleashed/ArchiveRecord.scala
```diff
@@ -18,16 +18,22 @@
...
 import org.apache.spark.SerializableWritable
 import org.archive.io.arc.ARCRecord
 import org.archive.io.warc.WARCRecord
 import org.archive.util.ArchiveUtils
+import scala.util.Try
+import org.apache.commons.httpclient.{Header, HttpParser, StatusLine}
 
 /** Trait for a record in a web archive. */
 trait ArchiveRecord extends Serializable {
+  /** Returns the filename containing the Archive Records */
```
ruebot: Missing fullstop.

src/main/scala/io/archivesunleashed/ArchiveRecord.scala
```diff
@@ -64,6 +74,7 @@ class ArchiveRecordImpl(r: SerializableWritable[ArchiveRecordWritable]) extends
   var arcRecord: ARCRecord = null
   var warcRecord: WARCRecord = null
   // scalastyle:on null
+  var headerResponseFormat: String = "US-ASCII"
```
ruebot: What's the rationale for `US-ASCII`?

ruebot:
```
   Historically, HTTP has allowed field content with text in the
   ISO-8859-1 charset [ISO-8859-1], supporting other charsets only
   through use of [RFC2047] encoding.  In practice, most HTTP header
   field values use only a subset of the US-ASCII charset [USASCII].
   Newly defined header fields SHOULD limit their field values to
   US-ASCII octets.  A recipient SHOULD treat other octets in field
   content (obs-text) as opaque data.
```

I suppose it this? https://tools.ietf.org/html/rfc7230#section-3.2.4

src/main/scala/io/archivesunleashed/ArchiveRecord.scala
greebie: Done.  Just to confirm Resourcename (to mimic Filename) and not ResourceName (because resource name are two separate words)? I'm okay with either.

src/main/scala/io/archivesunleashed/ArchiveRecord.scala
greebie: Yes - this part was a mimic of @dportabella 's suggestion. Seemed weird to me at first as well.

src/main/scala/io/archivesunleashed/ArchiveRecord.scala
greebie: Current commit uses Resourcename.  I think that's fine - I'm probably just overthinking it.

ianmilligan1: I think `Resourcename` works FWIW

greebie:
I would like to run one more test on this minus the `.keepValidRecords()`.  Seems like there's potential for that to break.

ianmilligan1:
So just i.e.

```scala
import io.archivesunleashed._

RecordLoader.loadArchives("/tuna1/scratch/i2milligan/warcs.archive-it.org/cgi-bin/getarcs.pl/*.gz", sc)
  .map(r => (r.getFilename, r.getHttpStatus, r.getCrawlDate, r.getDomain, r.getUrl, RemoveHTML(r.getContentString)))
  .saveAsTextFile("/tuna1/scratch/i2milligan/results/plain-text-http")
```

I have a large body of WARCs to test on, so if that's what you mean I can test it here too @greebie

greebie:
Yes.  Would like to see some statuses besides 200.  :)  Also if the header is munged, I'd like to be able to cover for it.

ianmilligan1:
OK. Let me run it at scale minus the plain text, so we can easily look around at status codes too.

ianmilligan1:
Tested it without `.keepValidPages()` and all seems to work.

Am seeing other response codes - i.e. 404s, 301s, etc. i.e.

```
(file:/tuna1/scratch/i2milligan/warcs.archive-it.org/cgi-bin/getarcs.pl/ARCHIVEIT-5421-MONTHLY-3224-20150514172923428-00000-wbgrp-crawl064.us.archive.org-6442.warc.gz,301,20150514,www.ontario.ca,http://www.ontario.ca/dgr)
(file:/tuna1/scratch/i2milligan/warcs.archive-it.org/cgi-bin/getarcs.pl/ARCHIVEIT-5421-MONTHLY-3224-20150514172923428-00000-wbgrp-crawl064.us.archive.org-6442.warc.gz,200,20150514,www1.toronto.ca,http://www1.toronto.ca/wps/portal/contentonly?vgnextoid=68a618b06a1aa410VgnVCM10000071d60f89RCRD)
(file:/tuna1/scratch/i2milligan/warcs.archive-it.org/cgi-bin/getarcs.pl/ARCHIVEIT-5421-MONTHLY-3224-20150514172923428-00000-wbgrp-crawl064.us.archive.org-6442.warc.gz,200,20150514,www.ontario.ca,http://www.ontario.ca/travel-and-recreation/about-ontario?utm_source=shortlinks&utm_medium=web&utm_campaign=dgr)
(file:/tuna1/scratch/i2milligan/warcs.archive-it.org/cgi-bin/getarcs.pl/ARCHIVEIT-5421-MONTHLY-3224-20150514172923428-00000-wbgrp-crawl064.us.archive.org-6442.warc.gz,404,20150514,www.ontario.ca,http://www.ontario.ca/travel-and-recreation/Msxml2.XMLHTTP)
```

greebie:
Awesome!  I think doing a status code visualisation would make a decent Medium post.  That might be my december project.

greebie:
More relevant to #260 but related to why coverage shrunk in this PR - I think we need test cases for ArchiveRecords that are neither WARC or ARC Format.  The java takes care of the error handling I think, but codecov notices that we did not test for that here in ArchiveRecord for some reason. I think this is something for a different PR, however.

src/test/scala/io/archivesunleashed/ArchiveRecordTest.scala
```diff
@@ -44,10 +45,68 @@ class ArchiveRecordTest extends FunSuite with BeforeAndAfter {
...
+  test("Crawl Dates") {
+    val textSampleArc = RecordLoader.loadArchives(arcPath, sc)
+     .map(x => x.getCrawlDate).take(3)
+    val textSampleWarc = RecordLoader.loadArchives(warcPath, sc)
+     .map(x => x.getCrawlDate).take(3)
+    assert(textSampleArc.deep == Array("20080430", "20080430", "20080430").deep)
+    assert(textSampleWarc.deep == Array("20080430", "20080430", "20080430").deep)
+  }
+
+  test("Domains") {
```
ruebot: There are a few extra test methods here. Are they scope for the issues posted in the original comment? Or do they cover other tickets?

greebie: No other tickets.  Basically, last PR I had good test coverage, but failed to test particular cases and my code passed Travis with bugs.  I decided to include these additional tests in case that happened again (it was unlikely, but I wanted this PR to go more smoothly). Since ArchiveRecord is used widely across the tests, I did not expect the tests to improve coverage (as per #260).

greebie:
> I think we have different interpretations of what Resource name is. This appears to be putting out the filename of the ARC/WARC. My understanding was that it was the resource name of the resource being parsed. i.e. `foo.jpg` or `index.html`. If we're resolving #164 with this (my bad, I should I have read that more closely when first reviewing), then it _should not_ be `Resourcename`. It should be `ArchiveFilename` or something similar that is more explicit what it is.

I think archiveFilename works.  The issue is that because the filename comes from the `ArcRecord` and `WarcRecord` classes it technically could be a url.  IIPC calls it "getReaderIdentifier()"

greebie:
> We also need to decide if it is just the ARC/WARC filename name itself that the method returns, or the full path.

I suggest we keep the fullpaths but provide information accessing the filename in the docs. There's an example to extract just the filename in the tests.

PR #318, opened 2019-04-25
add image analysis w/ tensorflow

h324yang:
# JCDL2019 demo
Using AUT and SSD model w/ Tensorflow to do object detection analysis on web archives.
* * *
1. default setting is `standalone` mode, so need to set up master and slaves first.
2. run `detect.py` to get and store the object probabilities and the image byte strings.
3. run `extract_images.py` to get image files from the result of `step2`

Issue #317, opened 2019-04-24
Heap space!! java.lang.OutOfMemoryError: Java heap space at java.util.Arrays.copyOf(Arrays.java:3332)


ruebot:
**Describe the bug**

On a number FDLP and Stanford collections, we run into this space heap space error, and it kills the Spark job. Upon investigation, this does not seem to be a problem with a large ARC/WARC or large collection. This happens on small and large collections.

```
19/04/18 18:02:34 DEBUG BlockManager: Told master about block rdd_3_3
19/04/18 18:02:34 INFO ArchiveRecordInputFormat: Closed archive file file:/tuna1/scratch/nruest/auk_collection_testing/10689/warcs/ARCHIVEIT-10689-TEST-JOB677126-20180828211602861-00009.warc.gz
19/04/18 18:02:34 ERROR Executor: Exception in task 3.0 in stage 0.0 (TID 3)
java.lang.OutOfMemoryError: Java heap space
        at java.util.Arrays.copyOf(Arrays.java:3332)
        at java.lang.StringCoding.safeTrim(StringCoding.java:89)
        at java.lang.StringCoding.access$100(StringCoding.java:50)
        at java.lang.StringCoding$StringDecoder.decode(StringCoding.java:154)
        at java.lang.StringCoding.decode(StringCoding.java:193)
        at java.lang.StringCoding.decode(StringCoding.java:254)
        at java.lang.String.<init>(String.java:546)
        at java.lang.String.<init>(String.java:566)
        at io.archivesunleashed.ArchiveRecordImpl.<init>(ArchiveRecord.scala:117)
        at io.archivesunleashed.package$RecordLoader$$anonfun$loadArchives$2.apply(package.scala:69)
        at io.archivesunleashed.package$RecordLoader$$anonfun$loadArchives$2.apply(package.scala:69)
        at scala.collection.Iterator$$anon$11.next(Iterator.scala:410)
        at scala.collection.Iterator$$anon$13.hasNext(Iterator.scala:463)
        at org.apache.spark.serializer.SerializationStream.writeAll(Serializer.scala:139)
        at org.apache.spark.serializer.SerializerManager.dataSerializeStream(SerializerManager.scala:174)
        at org.apache.spark.storage.BlockManager$$anonfun$doPutIterator$1$$anonfun$apply$10.apply(BlockManager.scala:1203)
[...]

Issue #308, opened 2019-02-18
Strategy to deal with conflict between application and Spark distribution dependencies


jrwiebe:
There is a conflict between Tika's dependency on `commons-compress` and the version that is included in the Spark distribution, which under normal parameters causes calls to `DetectMimeTypeTika` from spark-shell to fail with a `java.lang.NoSuchMethodError`. For example, running the following code:
```
import io.archivesunleashed._
import io.archivesunleashed.matchbox._

val r = RecordLoader.loadArchives("/path/to/example.warc.gz", sc)

r.keepValidPages().map(r => DetectMimeTypeTika(r.getContentString)).take(5)
```
results in:
```
2019-02-18 16:39:07 ERROR Executor:91 - Exception in task 0.0 in stage 0.0 (TID 0)
java.lang.NoSuchMethodError: org.apache.commons.compress.archivers.ArchiveStreamFactory.detect(Ljava/io/InputStream;)Ljava/lang/String;
	at org.apache.tika.parser.pkg.ZipContainerDetector.detectArchiveFormat(ZipContainerDetector.java:160)
...
```

This is because current Spark distributions include old versions of commons-compress, and our code depends on version >= 1.14, which introduced the `ArchiveStreamFactory.detect()` method. Spark shell's classpath, which includes its `jars/` directory, takes precedence over our dependencies.

I initially resolved this by [adding an exclusion](https://github.com/archivesunleashed/aut/commit/1041d78025d2180a1f06a11714d1b13d3d0fe715) of poi-ooxml to the tika-parsers dependency in our POM, since poi-ooxml is the module that requires the newer commons-compress. This won't do, however, since we will want that module for detection of Microsoft Office formats.

A better solution is to prepend the correct commons-compress JAR to the spark-shell classpath with the `--driver-class-path` argument (i.e., the `spark.driver.extraClassPath` property). E.g, 
```
$ spark-shell --jars target/aut-0.17.1-SNAPSHOT-fatjar.jar --driver-memory 4G --driver-class-path /home/jrwiebe/.m2/repository/org/apache/commons/commons-compress/1.18/commons-compress-1.18.jar
```

[...]

Issue #295, opened 2018-11-30
Test aut with Apache Spark 2.4.0


ruebot:
[Spark 2.4.0](https://spark.apache.org/releases/spark-release-2-4-0.html) came out in November 2018. We should put `aut` a few tests with it.

1. Update `pom.xml` to use Apache Spark 2.4.0
2. Update `pom.xml` to use Scala 2.11.12
3. Test all examples [here](https://archivesunleashed.org/aut/)
4. Test [examples](https://github.com/archivesunleashed/archivesunleashed.org/issues?q=is%3Aissue+is%3Aopen+label%3Arelease-documentation) slated for 0.18.0 release
5. Test with AUK

Did I miss anything?

ruebot:
Branch: https://github.com/archivesunleashed/aut/tree/spark-2.4.0

ruebot:
That's great! Thank you!!

I'll test again with AUK. I did it on Monday, and didn't [this](https://github.com/apache/spark/pull/22919). Trying to figure out if that made it into the release or not. Because the release notes have it as a known issue, but it worked on Monday. Maybe it's more of an FYI :man_shrugging: 

...anyway. Testing again!

ruebot:
Apache Spark 2.4.0 + aut 0.17.1-SNAPSHOT in AUK fails:

```
/home/nruest/bin/spark-2.4.0-bin-hadoop2.7/bin/spark-shell --master local\[10\] --driver-memory 30G --conf spark.network.timeout=10000000 --conf spark.executor.heartbeatInterval=600s --conf spark.driver.maxResultSize=4G --packages "io.archivesunleashed:aut:0.17.1-SNAPSHOT" -i /home/nruest/tmp/auk/75/3552/2/spark_jobs/3552.scala | tee /home/nruest/tmp/auk/75/3552/2/spark_jobs/3552.scala.log

:: problems summary ::
:::: WARNINGS
                [NOT FOUND  ] commons-configuration#commons-configuration;1.8!commons-configuration.jar (0ms)

        ==== local-m2-cache: tried

          file:/home/nruest/.m2/repository/commons-configuration/commons-configuration/1.8/commons-configuration-1.8.jar

                [NOT FOUND  ] org.apache.commons#commons-lang3;3.3.1!commons-lang3.jar (0ms)

        ==== local-m2-cache: tried

          file:/home/nruest/.m2/repository/org/apache/commons/commons-lang3/3.3.1/commons-lang3-3.3.1.jar

[...]

PR #321, opened 2019-07-04
Update to Spark 2.4.3 and update Tika to 1.20.

ruebot:
**GitHub issue(s)**:

- #295
- #308
- #286

Update to Spark 2.4.3 and update Tika to 1.20, and pulls in unfinished work by @jrwiebe and @borislin.

I had to tweak the language tests. I believe is because of the updated language detection in Tika, and that's why the test fixtures changed. Though, I'm not a 100% certain. So, definitely need a sanity check here.

# How should this be tested?

TravisCI should take care of the first bit.

@ianmilligan1 would you mind putting this through a bunch of examples?

- Pull down the branch
- `rm -rf ~/.m2/repository/* && mvn clean install`
- `rm -rf ~/.ivy2/* && spark-shell --packages "io.archivesunleashed:aut:0.17.1-SNAPSHOT"`
- Put it through its paces

# Additional Notes:

Once we get this in, and #318, might as well cut a 0.18.0 release. There's a fair bit done since the [last release](https://github.com/archivesunleashed/aut/compare/aut-0.17.0...master). It'd be nice to get #317 sorted, but that isn't a blocker.

@jrwiebe if you have some time, can you give this a sanity look too, since I'll built on a bit of what you were previously digging into.


src/main/scala/io/archivesunleashed/app/CommandLineApp.scala
```diff
@@ -178,9 +178,13 @@ class CommandLineApp(conf: CmdAppConf) {
 
   def save(d: Dataset[Row]): Unit = {
     if (!configuration.partition.isEmpty) {
-      d.coalesce(configuration.partition()).write.csv(saveTarget)
+      d.coalesce(configuration.partition()).write
+        .option("timestampFormat", "yyyy/MM/dd HH:mm:ss ZZ")
```
ruebot: FYI: These were added for the newer version of Tika (underlying libraries). Timestamp format is required.

PR #323, opened 2019-07-08
Remove Tweet utils.

### LINKED ISSUE #194: More complete Twitter Ingestion

### LINKED ISSUE #206: TweetUtilsTest.scala doesn't test Spark, only underlying json4s library
We should test the Tweet functionality with Spark instead of just the underlying library, that way things will break in tests, and not be silently broken for a few releases :wink:

### LINKED ISSUE #322: Remove Twitter/tweet analysis
@lintool @ianmilligan1 is there any major reason to keep the Tweet support around? It's not really fully fleshed out, and there are a bunch of dependencies we could clean out if we remove it.

ruebot:
**GitHub issue(s)**:

- Resolves #322
- Resolves #206
- Resolves #194

Remove all Tweet analysis functionality.

# How should this be tested?

- TravisCI

# Additional Notes:

Need to remove documentation on archivesunleashed.org for next release.


ianmilligan1:
Will test today or tomorrow, and I can stage the archivesunleashed.org PR as well.

PR #335, opened 2019-07-30
Make ArchiveRecord.getContentBytes consistent, Resolve #334

### LINKED ISSUE #334: Inconsistency in ArchiveRecord.getContentBytes
I noticed [`ArchiveRecord.getContentBytes`](https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/ArchiveRecord.scala) behaves differently on ARC and WARC records. For an ARC it return just the contents of the record, using `ArcRecordUtils.getBodyContent(arcRecord)`, but for a WARC it return the "raw" contents -- i.e., including HTTP headers -- using `WarcRecordUtils.getContent(warcRecord)`. I don't see a rationale for this difference; it seems unintentional.

@ruebot @lintool

ianmilligan1:
**GitHub issue(s)**:

#334

As noted in #334, @jrwiebe discovered that we are inconsistent on how we `getContentBytes` of ARC and WARC files. Currently, ARC files we just get the contents of the record (using `getBodyContent`) and WARC files we get everything (using `getContent`). The two should be consistent.

On the ticket itself, I discussed the various outputs, how they differ, and potential solutions. I think it is critical that we normalize the two approaches. Given the existence of `RemoveHttpHeader`, I think we should have both just using `getContent`. If people just want the body content, they can remove the header (which is what we've been doing with WARC files). I can imagine lots of diverse use cases for HTTP header information so it's better to have it in there and then remove it.

# How should this be tested?

- TravisCI _should_ turn green. Note that I had to change the tests because now that we are returning headers in the ARC files, the tika and language counts are different. This is to be expected, to me at least.
- We should do some sanity checking on the outputs.

# Additional Notes:

We should probably foreground `RemoveHttpHeader` more in our documentation. 

# Interested parties

@ruebot @jrwiebe

### DIFF
--- a/src/main/scala/io/archivesunleashed/ArchiveRecord.scala
+++ b/src/main/scala/io/archivesunleashed/ArchiveRecord.scala
@@ -107,7 +107,7 @@ class ArchiveRecordImpl(r: SerializableWritable[ArchiveRecordWritable]) extends 
   val getContentBytes: Array[Byte] = {
     if (recordFormat == ArchiveRecordWritable.ArchiveFormat.ARC)
     {
-      ArcRecordUtils.getBodyContent(r.t.getRecord.asInstanceOf[ARCRecord])
+      ArcRecordUtils.getContent(r.t.getRecord.asInstanceOf[ARCRecord])
     } else {
       WarcRecordUtils.getContent(r.t.getRecord.asInstanceOf[WARCRecord])
     }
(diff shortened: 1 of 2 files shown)


ianmilligan1:
>I'm going to test with language and full text extraction. That the big stuff?

Yep! 

> Should I through RemoveHttpHeader into the scripts since we see the headers a fair bit in our results?

Good idea. Behaviour should now always be consistent between ARCs + WARCs. 

I'll open up a documentation ticket.

jrwiebe:
A-OK

Issue #343, opened 2019-08-13
Add method for unknown extensions in binary extractions


ruebot:
With the implementation of #302, #306, and #307, we will occasionally get binaries that are extracted, and do not have file extensions on them. We should create a method/helper account for this:

1. Try existing method
2. Try to guess file extension from MimeType (@jrwiebe is working on this in [get-extension](https://github.com/archivesunleashed/aut/tree/get-extension)
3. If both fail, use `UNKNOWN` or something else as the extension.

jrwiebe:
I noticed when I was filtering for URLs ending in ".doc" that I was getting a lot of non-doc formats (HTML and text formats). I think it's less likely there will be such incorrect file extensions for the other binary formats we're targeting, but if we want a generic algorithm for determining the extension, I'd cover the .doc case by switching steps 1 and 2. I'd also use the plural method `getExtensions` and if the extension returned by FilenameUtils is contained in this list, I'd select that one.

We're already getting the MimeType with Tika, which is the only expensive operation in this process.

My method:
```
  def getExt(mimeType: String, url: String): String = {
    val tikaExtensions = DetectMimeTypeTika.getExtensions(mimetype)
    var ext = "unknown"
    // Tika method
    if (tikaExtensions.size == 1) {
      ext = tikaExtensions(0).substring(1)
    } else {
      // FilenameUtils method
      val fnuExt = FilenameUtils.getExtension(url)
      if (fnuExt != null && !fnuExt.isEmpty) {
        // Reconcile Tika list and FilenameUtils extension
        if (tikaExtensions.size > 1) {
          if (tikaExtensions.contains(fnuExt)) {
            ext = fnuExt
          } else {
            ext = tikaExtensions(0).substring(1)
          }
        } else { // tikaExtensions.size == 0 && fnuExt exists
          ext = fnuExt
        }
      } // else => unknown
    }
    ext
  }
```

ruebot:
I have an idea since I noticed these methods as I'm hacking on #346 

https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/package.scala#L307-L313

https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/package.scala#L374-L380

[...]

PR #349, opened 2019-08-17
Add method for determining binary file extension

jrwiebe:
**GitHub issue(s)**:

* #343 

This PR implements the strategy described in the discussion of the above issue to get an extension for a file described by a URL and a MIME type. It creates a `GetExtensionMime` object in the matchbox.

This PR also removes most of the filtering by URL from the image, audio, video, presentation, spreadsheet, and word processor document extraction methods, since these were returning false positives. (CSV and TSV files are a special case, since Tika detects them as "text/plain" based on content.)

Finally, I have inserted `toLowerCase` into the `getUrl.endsWith()` filter tests, which could possibly bring in some more CSV and TSV files 

# How should this be tested?

Test by running something like the following script first on a build of `master`, then modify the output path and do the same on a build of `get-extension`. Depending on your input there may or may not be a difference between the sets of files that are extracted. If there is, the second run should have fewer files of all types except images, due to misidentification of files by URL in the first run (i.e., false positives), and they should all have extensions. Because `extractImageDetailsDF` was using the MIME type stored in the archive record and not the detected version, the first run might produce fewer image files than the second (i.e., `master` was producing false negatives); the `master` version's reliance on the URL extension could also produce false positives. Because we 

(Tip: You can use the MD5 hash in the filenames to identify files with the same content.)
```
import io.archivesunleashed._

val warcs_path = "/home/jrwiebe/warcs/cpp10/*.gz"
val output_path = "/tuna1/scratch/jrwiebe/get-extension-test/master/"

val df_ss = RecordLoader.loadArchives(warcs_path, sc).extractSpreadsheetDetailsDF();
val res_ss = df_ss.select($"bytes", $"extension").saveToDisk("bytes", output_path+"spreadsheet", "extension")

val df_pp = RecordLoader.loadArchives(warcs_path, sc).extractPresentationProgramDetailsDF();
val res_pp = df_pp.select($"bytes", $"extension").saveToDisk("bytes", output_path+"presentation", "extension")

val df_word = RecordLoader.loadArchives(warcs_path, sc).extractWordProcessorDetailsDF();
val res_word = df_word.select($"bytes", $"extension").saveToDisk("bytes", output_path+"document", "extension")

val df_img = RecordLoader.loadArchives(warcs_path, sc).extractImageDetailsDF();
val res_img = df_img.select($"bytes", $"extension").saveToDisk("bytes", output_path+"image", "extension")

val df_aud = RecordLoader.loadArchives(warcs_path, sc).extractAudioDetailsDF();
val res_aud = df_aud.select($"bytes", $"extension").saveToDisk("bytes", output_path+"audio", "extension")

val df_vid = RecordLoader.loadArchives(warcs_path, sc).extractWordProcessorDetailsDF();
val res_vid = df_vid.select($"bytes", $"extension").saveToDisk("bytes", output_path+"video", "extension")

sys.exit
```

[Here are my results](https://gist.github.com/jrwiebe/dfbe6c4673167219b088bc22f5171158). For the document, spreadsheet, and presentation files I confirmed that files missing from the second run were files that had been misidentified in the first run (master branch).

Admittedly mine wasn't a complete test, since it doesn't show how `GetExtensionMime` would handle a file with the wrong extension in the URL. @ruebot, since the tests you created recently reference actual files on your web server, maybe you could add a couple? To demonstrate how the method does work, see:
```
scala> import io.archivesunleashed.matchbox._

scala> GetExtensionMime("http://ruebot.net/misnameddoc.exe", "application/vnd.openxmlformats-officedocument.wordprocessingml.document")
19/08/16 12:42:02 WARN PDFParser: J2KImageReader not loaded. JPEG2000 files will not be processed.
See https://pdfbox.apache.org/2.0/dependencies.html#jai-image-io
for optional dependencies.

19/08/16 12:42:02 WARN TesseractOCRParser: Tesseract OCR is installed and will be automatically applied to image files unless
you've excluded the TesseractOCRParser from the default parser.
Tesseract may dramatically slow down content extraction (TIKA-2359).
As of Tika 1.15 (and prior versions), Tesseract is automatically called.
In future versions of Tika, users may need to turn the TesseractOCRParser on via TikaConfig.
19/08/16 12:42:02 WARN SQLite3Parser: org.xerial's sqlite-jdbc is not loaded.
Please provide the jar on your classpath to parse sqlite files.
See tika-parsers/pom.xml for the correct version.
res0: String = docx

scala> GetExtensionMime("http://ruebot.net/this_is_an_mp3", "audio/mpeg")
res1: String = mpga
```
(`mpga` might) be unexpected here, but it is the first in the list of extensions associated with the MIME type `audio/mpeg`. Oh, well.)

# Additional notes
1) Part of the reason for false positives was that [blocks like this](https://github.com/archivesunleashed/aut/blob/c824ad814f1c6ebc758597d3ba6bfa57c4c3e43f/src/main/scala/io/archivesunleashed/package.scala#L327-L336) should be ANDs, not ORs.

2) I left the `!r.getUrl.endsWith("robots.txt")` condition in `keepImages`, because removing it caused a few files to be found that looked like GIFs and JPEGs, but which were named `/robots.txt` and which were incomplete, causing `saveImageToDisk` to fail with a `java.io.EOFException`.

3) I don't know if we need to be using `saveImageToDisk`. We could simply use `saveToDisk`. This PR adds the "extension" (and "file") field to the DF returned by `extractImageDetailsDF`, so using the later save method is now an option.


jrwiebe:
If we deprecated `saveImageToDisk` in favour of simply using `saveToDisk`, we could safely remove the `robots.txt` check, since the generic save message does not read the binary bytes to ensure they represent a complete, well-formed file. 

This isn't a big deal. I like removing the robots check to make the code more elegant. And theoretically a URL ending with "robots.txt" could actually be an image – though this is unlikely.

jrwiebe:
@ruebot How about `this_is_a_gif` (no extension) and `this_is_a_jpeg.mp3` (JPEG).

jrwiebe:
@ruebot Yes

jrwiebe:
@ruebot Would you mind replacing `this_is_a_jpeg.mp3` with an actual JPEG file? I wanted to test the case where the Tika extension and the FilenameUtils one differ.

jrwiebe:
I didn't get that in my test, but my WARCs might contain fewer files. Try throwing a `file.close()` after [this line](https://github.com/archivesunleashed/aut/blob/get-extension/src/main/scala/io/archivesunleashed/df/package.scala#L65).

jrwiebe:
@ruebot I fixed the tests, but if you want to tweak them that's fine. I think we're ready to go.

Issue #306, opened 2019-01-31
Video binary object extraction


ruebot:
Using the [image extraction process](https://github.com/archivesunleashed/aut/pull/234) as a basis, our next set of binary object extractions will be documents. This issue is meant to focus specially on video objects. 

There may be a some tweaks to this depending on the outcome of #298.

ruebot:
Ok, I have a basic framework setup in [the branch](https://github.com/archivesunleashed/aut/tree/issue-306-307).

Pull down the branch, and do something along these lines (I had memory issues, so I did my whole Spark config thing):

```
rm -rf ~/.m2/repository/* && mvn clean install && rm -rf ~/.ivy2/* && ~/bin/spark-2.4.3-bin-hadoop2.7/bin/spark-shell --master local\[10\] --driver-memory 35g --conf spark.network.timeout=10000000 --conf spark.executor.heartbeatInterval=600s --conf spark.driver.maxResultSize=4g --conf spark.serializer=org.apache.spark.serializer.KryoSerializer --conf spark.shuffle.compress=true --conf spark.rdd.compress=true --packages io.archivesunleashed:aut:0.17.1-SNAPSHOT -i ~/306-pdf-audio-video-extract.scala
```

**306-pdf-audio-video-extract.scala**
```scala
import io.archivesunleashed._
import io.archivesunleashed.df._

val df_pdf = RecordLoader.loadArchives("/home/nruest/Projects/au/sample-data/geocites/1/*gz", sc).extractPDFDetailsDF();
val res_pdf = df_pdf.select($"bytes", $"extension").saveToDisk("bytes", "/home/nruest/Projects/au/sample-data/306-307-test/pdf", "extension")

val df_audio = RecordLoader.loadArchives("/home/nruest/Projects/au/sample-data/geocites/1/*gz", sc).extractAudioDetailsDF();
val res_audio = df_audio.select($"bytes", $"extension").saveToDisk("bytes", "/home/nruest/Projects/au/sample-data/306-307-test/audio", "extension")

val df_video = RecordLoader.loadArchives("/home/nruest/Projects/au/sample-data/geocites/1/*gz", sc).extractVideoDetailsDF();
val res_video = df_video.select($"bytes", $"extension").saveToDisk("bytes", "/home/nruest/Projects/au/sample-data/306-307-test/video", "extension")

sys.exit
```

[...]

Issue #304, opened 2019-01-31
Doc binary object extraction


ruebot:
Using the [image extraction process](https://github.com/archivesunleashed/aut/pull/234) as a basis, our next set of binary object extractions will be documents. This issue is meant to focus specially on `doc`  `docx`, and `odt`.

There may be a some tweaks to this depending on the outcome of #298.

Issue #303, opened 2019-01-31
Spreadsheet binary object extraction


ruebot:
Using the [image extraction process](https://github.com/archivesunleashed/aut/pull/234) as a basis, our next set of binary object extractions will be documents. This issue is meant to focus specially on `xls`  `xlsx`, `ods`, and `csv`.

There may be a some tweaks to this depending on the outcome of #298.

jrwiebe:
Putting this here for my reference; feedback is welcome.

These are the spreadsheet MIME types [Tika will identify](https://tika.apache.org/1.20/formats.html). The [Mozilla MIME type list](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Complete_list_of_MIME_types) was also consulted. Unless there are objections I think I'll extract all of these, including templates and MS Works spreadsheets. 

### Excel
application/vnd.ms-excel
application/vnd.ms-excel.workspace.3
application/vnd.ms-excel.workspace.4
application/vnd.ms-excel.sheet.2
application/vnd.ms-excel.sheet.3
application/vnd.ms-excel.sheet.4
application/vnd.ms-excel.addin.macroenabled.12
application/vnd.ms-excel.sheet.binary.macroenabled.12
application/vnd.ms-excel.sheet.macroenabled.12
application/vnd.ms-excel.template.macroenabled.12
application/vnd.ms-spreadsheetml

### Open Office
application/vnd.openxmlformats-officedocument.spreadsheetml.template
application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
application/x-vnd.oasis.opendocument.spreadsheet-template
application/vnd.oasis.opendocument.spreadsheet-template
application/vnd.oasis.opendocument.spreadsheet
application/x-vnd.oasis.opendocument.spreadsheet

### Other
application/x-tika-msworks-spreadsheet

### CSV

Currently Tika only detects CSV if the parser is given a filename with the extension "CSV", although byte-based detection might be coming ([TIKA-2826](https://issues.apache.org/jira/browse/TIKA-2826)). I'll detect CSV by looking at the URL extension and checking if `getMimeType() == "text/csv"`.

jrwiebe:
### MIME type references for #303, #304, #305, #306, #307:
* [Tika: Full list of supported formats](https://tika.apache.org/1.20/formats.html#Full_list_of_Supported_Formats)
[...]

Issue #302, opened 2019-01-31
PDF binary object extraction


ruebot:
Using the [image extraction process](https://github.com/archivesunleashed/aut/pull/234) as a basis, our next set of binary object extractions will be documents. This issue is meant to focus specially on PDFs.

There may be a some tweaks to this depending on the outcome of #298.

ruebot:
So, I think I have it working now building off of @jrwiebe's [extract-pdf branch](https://github.com/archivesunleashed/aut/tree/extract-pdf). I tested two scripts -- PDF binary extraction, and PDF details data frame -- on 878 GeoCities WARCs on tuna.

## PDF extraction
<hr />

**Script**
```scala
import io.archivesunleashed._
import io.archivesunleashed.df._

val df = RecordLoader.loadArchives("/tuna1/scratch/nruest/geocites/warcs/9/*.gz", sc).extractPDFDetailsDF();  
val res = df.select($"bytes").saveToDisk("bytes", "/tuna1/scratch/nruest/geocites/pdfs/9/aut-302-test", "pdf")
sys.exit
```

**Job**
```bash
$ time /home/ruestn/spark-2.4.3-bin-hadoop2.7/bin/spark-shell --master local[75] --driver-memory 300g --conf spark.network.timeout=10000000 --conf spark.executor.heartbeatInterval=600s --conf spark.driver.maxResultSize=4g --conf spark.serializer=org.apache.spark.serializer.KryoSerializer --conf spark.shuffle.compress=true --conf spark.rdd.compress=true --packages "io.archivesunleashed:aut:0.17.1-SNAPSHOT" -i /home/ruestn/aut-302-pdf-extraction/aut-302-pdf-extraction.scala 2>&1 | tee /home/ruestn/aut-302-pdf-extraction/logs/set-09.log
```

**Results**

```
$ ls | wc -l
144757

$ du -sh
23G	.
```

Example output: https://www.dropbox.com/s/iwic5pwozikye5i/aut-302-test-925e8751447c08f2fbdf175e9560df7a.pdf

## Data frame to csv
<hr />

**Script**
```scala
import io.archivesunleashed._
import io.archivesunleashed.df._

val df = RecordLoader.loadArchives("/tuna1/scratch/nruest/geocites/warcs/9/*gz", sc).extractPDFDetailsDF();
df.select($"url", $"mime_type", $"md5").orderBy(desc("md5")).write.csv("/home/ruestn/aut-302-pdf-extraction/df/9")

[...]

PR #350, opened 2019-08-20
Add binary extraction DataFrames to PySpark.

ruebot:
**GitHub issue(s)**:

- #190
- #259
- #302
- #303
- #304
- #305
- #306
- #307

Add binary extraction DataFrames to PySpark.

# How should this be tested?

- TravisCI
- Fire up a Jupyter Notebook:
```
$ PYSPARK_DRIVER_PYTHON=jupyter PYSPARK_DRIVER_PYTHON_OPTS=notebook ~/bin/spark-2.4.3-bin-hadoop2.7/bin/pyspark --jars /home/nruest/Projects/au/aut/target/aut-0.17.1-SNAPSHOT-fatjar.jar --driver-class-path /home/nruest/Projects/au/aut/target/aut-0.17.1-SNAPSHOT-fatjar.jar --py-files /home/nruest/Projects/au/aut/target/aut.zip
```

Then do this for each of the additions, and make sure it works:

```
from aut import *

archive = WebArchive(sc, sqlContext, "/home/nruest/Projects/au/aut/src/test/resources/warc/")

df = archive.spreadsheets()
df.printSchema()
```

Screenshot from 2019-08-20 13-25-51

# Additional Notes:

Let's see what happens with the test coverage. I assume I'm going to have to add some things to `src/test/scala/io/archivesunleashed/df/DataFrameLoaderTest.scala`

### DIFF
--- a/src/main/python/aut/common.py
+++ b/src/main/python/aut/common.py
@@ -13,3 +13,30 @@ class WebArchive:
 
     def links(self):
         return DataFrame(self.loader.extractHyperlinks(self.path), self.sqlContext)
+
+    def images(self):
+        return DataFrame(self.loader.extractImages(self.path), self.sqlContext)
+
+    def image_links(self):
+        return DataFrame(self.loader.extractImageLinks(self.path), self.sqlContext)
+
+    def pdfs(self):
+        return DataFrame(self.loader.extractPDFs(self.path), self.sqlContext)
+
+    def audio(self):
+        return DataFrame(self.loader.extractAudio(self.path), self.sqlContext)
+
+    def video(self):
+        return DataFrame(self.loader.extractVideo(self.path), self.sqlContext)
+
+    def spreadsheets(self):
+        return DataFrame(self.loader.extractSpreadsheets(self.path), self.sqlContext)
+
+    def presentation_program(self):
+        return DataFrame(self.loader.extractPresentationProgram(self.path), self.sqlContext)
+
+    def word_processor(self):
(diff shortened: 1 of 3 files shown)


src/main/python/aut/common.py
```diff
@@ -13,3 +13,30 @@ def pages(self):
 
     def links(self):
         return DataFrame(self.loader.extractHyperlinks(self.path), self.sqlContext)
+
+    def images(self):
+        return DataFrame(self.loader.extractImages(self.path), self.sqlContext)
+
+    def image_links(self):
```
lintool: Sorry for arriving late at the party - some of these DataFrames are singular, e.g., `presentation_program`, while others are plural, e.g., `images`... perhaps we might standardize?

Has this been discussed already?

Also, any issues in calling these what they are? word, powerpoint, excel, etc.?

If so, how about documents, presentations, and spreadsheets?

lintool:
Comments about naming df's if it's not too late...

src/main/python/aut/common.py
ruebot: > Has this been discussed already?

Pluralization, no. Hasn't been discussion. I just went with what made sense grammatically. They _should_ all be plural-ish.  

> Also, any issues in calling these what they are? word, powerpoint, excel, etc.?

I called them what the Wikipedia entry for the type of file called them since they aren't just the Microsoft products.

I thought about doing that with documents, presentations, and spreadsheets, but documents can be misleading with text files. So, I went with appear to be the canonical terms.

Issue #238, opened 2018-05-24
Dataframe Code Request: Finding Hyperlinks within Collection on Pages with Certain Keyword


ianmilligan1:
# Use Case

I am interested in finding what pages particular organizations link to when they contain certain keywords. 

In this case, I am curious to see where the Liberal Party of Canada (liberal.ca) has linked to from pages that contain the words "Keystone". Do they link to news sources? Do they link to activist groups? Do they link to opposition parties? Let's find out.

## Data

In this case we need to do the following:
* Find all pages within a collection of WARCs that contain the keyword `Keystone`. Let's make it case insensitive.
* Find all hyperlinks that are leaving the pages that contain the keyword `Keystone`.

## Desired Output

The desired output would be a CSV looking like:

```
domain, URL, crawl date, origin page, destination page
```

## Question

What data frame query can we use to first filter down to pages containing a certain keyword, and then extract just links from them? Hopefully this is clear but do let me know if I can clarify further.

ruebot:
@ianmilligan1 do you still need help on this one?

ianmilligan1:
I _think_ the problem here is that we don’t have text as part of our dataframe implementation yet.

ruebot:
@SinghGursimran here's one for you.

SinghGursimran:
import io.archivesunleashed._
import io.archivesunleashed.df._

val result = udf((vs: Seq[Any]) => vs(0).toString.split(",")(1))

val df= RecordLoader.loadArchives("src/test/resources/warc/example.warc.gz", sc).extractValidPagesDF()
  .select(RemovePrefixWWW(ExtractDomain($"url")).as("Domain"),$"url".as("url"),$"crawl_date",explode_outer(ExtractLinks($"url",$"content")).as("link"))
  .filter($"content".contains("keyNote"))

  df.select($"url",$"Domain",$"crawl_date",result(array($"link")).as("destination_page"))
  .write
  .option("header","true")
  .csv("filtered_results/")

I have written this script to answer the above query. It works but things can be done in a better way if I create a separate function in the code specifically for this query. I am not sure whether that's a good idea considering there can be many random queries. @ruebot @lintool @ianmilligan1

ruebot:
We have to import `matchbox` on that for `ExtractLinks`, and then we hit:

```scala
import io.archivesunleashed._
import io.archivesunleashed.df._
import io.archivesunleashed.matchbox._

[...]

PR #377, opened 2019-11-12
Finding Hyperlinks within Collection on Pages with Certain Keyword

SinghGursimran:
Extract hyperlinks within a collection filtered on pages containing a particular keyword (case insensitive) using df.

```scala

import io.archivesunleashed._

val result = udf((vs: Seq[Any]) => vs(0).toString.split(",")(1))

val df= RecordLoader.loadArchives("src/test/resources/warc/example.warc.gz", sc).extractValidPagesDF()
.select(     RemovePrefixWWW(ExtractDomain($"url")).as("Domain"),
  	        $"url".as("url"),
  	        $"crawl_date",
  	        explode_outer(ExtractLinks($"url",$"content")).as("link")
  	  )
.filter(lower($"content").contains("internet")) //filtered on keyword "internet"

df.select($"url",$"Domain",$"crawl_date",result(array($"link")).as("destination_page"))
.write
.option("header","true")
.csv("filtered_results/")

```
#238 

Returns a csv file with URL, Domain, crawl_date, and destination_page.

### DIFF
--- a/src/main/scala/io/archivesunleashed/df/package.scala
+++ b/src/main/scala/io/archivesunleashed/df/package.scala
@@ -16,7 +16,7 @@
 package io.archivesunleashed
 
 import org.apache.commons.io.IOUtils
-import io.archivesunleashed.matchbox.{ComputeMD5, ExtractDomain, RemoveHTML}
+import io.archivesunleashed.matchbox.{ComputeMD5, ExtractDomain, RemoveHTML, ExtractLinks}
 import org.apache.spark.sql.functions.udf
 import org.apache.spark.sql.DataFrame
 import java.io.ByteArrayInputStream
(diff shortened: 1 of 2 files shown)


lintool:
hey @ruebot is this something that should be encoded in a test case while we're at it?

lintool:
@ruebot add a separate test case that explicitly include filtering? Maybe not. I dunno.

ianmilligan1:
@ruebot Looks perfect to me - and I love the sample results with "keystone" leading to David Suzuki. Thanks @SinghGursimran, great work.

SinghGursimran:
> Yeah, we could add a new test that add `filter` or just add a `filter` to https://github.com/archivesunleashed/aut/blob/e32ae17c55740c6c6adc177d42690b39fa6321dd/src/test/scala/io/archivesunleashed/matchbox/ExtractLinksTest.scala if that's what you're thinking.
> 
> ...and if that's the case, @SinghGursimran, want to update the PR with an updated test?

Issue #366, opened 2019-10-20
Align DataFrame boilerplate in Python and Scala


lintool:
Currently, the Scala DF boilerplate is something like:

```
RecordLoader.loadArchives("src/test/resources/warc/example.warc.gz", sc).extractValidPagesDF()
  .select($"Url")
  .show(20, false)
```

And Python:

```
WebArchive(sc, sqlContext, "src/test/resources/warc/example.warc.gz").pages() \
    .select("url") \
    .show(20, False)
```

It would make sense to align?

So Python would look like:

```
RecordLoader.loadArchives("src/test/resources/warc/example.warc.gz", sc, sqlContext).pages() \
    .select("url") \
    .show(20, False)
```
And from the Scala end, let's just change `extractValidPagesDF()` to `pages()` to match the Python end? This runs the slight risk of confusion with RDD operations, but I think the risk is minimal.

@ianmilligan1 @ruebot thoughts?

ruebot:
That was the general consensus in https://github.com/archivesunleashed/aut/issues/231

lintool:
For references, where the current DF extractors are defined:
https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/package.scala#L102

And the Python connectors:
https://github.com/archivesunleashed/aut/blob/master/src/main/python/aut/common.py

Issue #371, opened 2019-11-05
Convert RecordLoader.loadArchives to a Spark Data Source


ruebot:
Since we're pivoting to full DataFrame support (#223, #190), we should convert/migrate `RecordLoader.loadArchives`, and any other related functions to a [Spark Data Source](https://spark.apache.org/docs/latest/sql-data-sources.html). That way we could do things like:

```scala
spark.read.format("webArchive")
  .option("mode", "FAILFAST")
  .option("inferSchema", "true")
  .option("/path/to/files")
  .schema(someSchema)
  .load()
```

Then, we _could_, (since it's an open issue #147) write WARCs that way too? :man_shrugging: 

```scala
spark.write.format("webArchive")
  .option("mode", "OVERWRITE")
  .option("/path/to/files")
  .save()
```

These are the Spark core data sources:

- CSV
- JSON
- Parquet
- ORC
- JDBC/ODBC
- Plain-text
- Avro

Community implemented data sources:

- Cassandra
- HBase
- MongoDB
- AWS Redshift
- XML

ruebot:
Some helpful links:

- https://mapr.com/blog/spark-data-source-api-extending-our-spark-sql-query-engine/
- http://blog.madhukaraphatak.com/spark-datasource-v2-part-3/
- https://databricks.com/session/how-to-connect-spark-to-your-own-datasource

sepastian:
https://jaceklaskowski.gitbooks.io/mastering-spark-sql/spark-sql-DataSourceV2.html
https://github.com/spirom/spark-data-sources

ruebot:
Cassandra [example](https://github.com/datastax/spark-cassandra-connector/blob/master/connector/src/main/scala/com/datastax/spark/connector/datasource/CassandraCatalog.scala)

ruebot:
I'm thinking this is out of scope for this project given the work being done on #494 now. So, I'm going to close it as won't fix.

PR #379, opened 2019-11-14
Rename DF functions to be consistent with Python DF functions.

ruebot:
**GitHub issue(s)**: #366 

Rename DF functions to be consistent with Python DF functions.

# How should this be tested?

- TravisCI (unit tests) should take care of it.

# Additional Notes:

1. @lintool the only part of the issue I haven't resolved here is your proposal to move `WebArchive` to `RecordLoader.loadArchives` on the Python side. I don't believe we can have `RecordLoader.loadArchives` as an actual class name. So, should we change things in the other direction or both directions? I've been naming a `val` `warcs` a lot in my example. What if we did something like that? Or `webArchives` or `webarchives` ? Once we settle on something, I'll the PR out of draft.

2. There will be a **TONNE** of documentation updates to do. If I could get a hand there, that'd be super helpful. @SamFritz @ianmilligan1 that of any interest?


ianmilligan1:
👀 

Looks like a great PR in progress, and agreed, the documentation will be heavy-ish lifting. I can help out if you give me discrete chunks of work (between parental leave, grant writing, admin, and some unfortunate university admin work coming up, my time is a bit limited, but many hands makes for light work).

lintool:
@rubot I'm not wedded to perfect alignment between Scala DF and Pyspark DF at this point... perhaps we should just wait for #371?

I would rather do the refactoring incrementally... i.e., let's get this merged, start fixing docs, etc.

ianmilligan1:
Great! @ruebot and I can maybe work together or the docs? When I have time I can start a branch or vice versa.

ianmilligan1:
Sure, sounds good on updating the PR and going fro, there! I might be able to tackle some of this tomorrow, kids schedule permitting.

PR #380, opened 2019-11-17
Matchbox utilities to DataFrames

SinghGursimran:
Extended matchbox utilities to df

Issue: #223

For Testing:

```scala
import io.archivesunleashed.df._
import io.archivesunleashed._

RecordLoader.loadArchives("./ARCHIVEIT-227-QUARTERLY-XUGECV-20091218231727-00039-crawling06.us.archive.org-8091.warc.gz", sc)
						.pages()
						.select(GetExtensionMime(($"url"),($"mime_type_web_server")).as("extension"))
						.show(20,false)

RecordLoader.loadArchives("./ARCHIVEIT-227-QUARTERLY-XUGECV-20091218231727-00039-crawling06.us.archive.org-8091.warc.gz", sc)
						.pages()
						.select($"url".as("url"), explode_outer(ExtractImageLinks(($"url"), ($"content"))).as("imageLinks"))
						.show(20,false)

RecordLoader.loadArchives("./ARCHIVEIT-227-QUARTERLY-XUGECV-20091218231727-00039-crawling06.us.archive.org-8091.warc.gz", sc)
						.pages()
						.select(ComputeMD5DF(($"content")).as("MD5Hash"))
						.show(20,false)

RecordLoader.loadArchives("./ARCHIVEIT-227-QUARTERLY-XUGECV-20091218231727-00039-crawling06.us.archive.org-8091.warc.gz", sc)
						.pages()
						.select(ComputeSHA1(($"content")).as("SHA1Hash"))
						.show(20,false)
```

PR #382, opened 2019-11-20
Extract popular images - Data Frame implementation

SinghGursimran:
Extract popular images - Data Frame implementation

#380  

For Testing:
```scala

import io.archivesunleashed._

val df = RecordLoader.loadArchives("./ARCHIVEIT-227-QUARTERLY-XUGECV-20091218231727-00039-crawling06.us.archive.org-8091.warc.gz",sc)
					 .images()	  

ExtractPopularImages(df,10,30,30).show()

```


SinghGursimran:
Scala doesn't support function overloading with default arguments. For the RDD implementation, minWidth and minHeight arguments were optional. For the current data frame implementation, they are necessary. If it is required to be kept as optional, I can 
1. Shift df implementation to a new object (ExtractPopularImagesDF) 
      OR
2. give a new name to the method. (That would require calling method name along with the object at the time of implementation)

ianmilligan1:
>A new method, ExtractPopularImagesDF makes sense to me. @ianmilligan1 @lintool work for you, or y'all prefer another path?

That makes sense to me too!

lintool:
Let's use a different convention for "end-to-end" functionalities. One option would be to have all UDFs be verb phrases, e.g., `ExtractX`, and "end-to-end" functionalities be noun phrases. So this would be `PopularImagesExtractor`. That way it'll be clear on what to use in what context.

lintool:
Yes, if you like my suggestion of nouns vs. verbs.

I.e., UDFs are verbs, "do this".
"Assemblies" are thing-doers.

SinghGursimran:
@ruebot 
Ya! I will change accordingly.
Regarding tests, there's an orderBy() function in implementation. If the value of key is same it orders randomly. If the count of two image URLs is same, the order of Data Frame row might change on running the code again. For the archive in resources, all images appear only once that is they all have count one. Adding a static test won't be feasible in such a case.

SinghGursimran:
> @SinghGursimran so, for the test. Can we assert other items in the DataFrame that is returned, that is not dependent on the order it returns in? `count` maybe? Or, combine `distinct` and `count` to hit the same thing each time?

Actually, for the archive available in the resources, the count is 1 for each data entry in the row.
The sequence of the whole row changes for the same count on subsequent running of the code. 
But, yes, I can keep count as the factor because it is already 1. Shall I do that?

Issue #223, opened 2018-05-15
Migration of all RDD functionality over to DataFrames


lintool:
We need to migrate all current RDD functionality over to DataFrames. This means porting all matchbox UDFs over to DF UDFs.

There are two possible ways to do this - we can simply take matchbox UDFs and wrap them, or rewrite them from scratch. I suggest we revisit one by one, which will give us an opportunity to refine the UDF we actually want.

For example, the current RDD matchbox [`ExtractDomain`](https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/matchbox/ExtractDomain.scala) is implemented a bit differently than the DF version we've been playing with... it, for example, strips the prefix `www`, whereas the RDD impl doesn't. I like the newer implementation better, but open to discussion.

Also, this is an issue we'll come across sooner or later:
https://stackoverflow.com/questions/33664991/spark-udf-initialization

I have a general question we need to look into from the performance perspective: what's the lifecycle status of a Spark DF UDF? In particular, if there's initialization like compiling regexp, we don't want to do that over and over again... we want to have an init stage?

@TitusAn let's start developing in parallel the DF versions of the apps in #222 and try and work this out?

ruebot:
@jrwiebe I have a note here from our call to "Go through the scala dir and identify all the functions that take in RDD and do not take in DF, then create tickets for JWb." Do you want me to do that granular of a level, or do you want to use this issue to take care of it?

jrwiebe:
This is fine.

ruebot:
@ianmilligan1 @lintool this look the basic inventory?

| rdd                        | data frame          |
|----------------------------|---------------------|
| `keepValidPages`           | `extractValidPages` |
|                            | `extractHyperlinks` |
|                            | `extractHyperlinks` |
|                            | `extractImages`     |
| `keepImages`               |                     |
| `keepMimeTypes`            |                     |
| `keepUrls`                 |                     |
| `keepUrlPatterns`          |                     |
| `keepDomains`              |                     |
| `keepLanguages`            |                     |
| `keepContent`              |                     |
| `discardMimeTypes`         |                     |
| `discardDate`              |                     |
| `discardUrls`              |                     |
| `discardUrlPatterns`       |                     |
| `discardDomains`           |                     |
| `discardContent`           |                     |
| `extractFromRecords`       |                     |
| `extractFromScrapeText`    |                     |
| `WriteGEXF`                |                     |
[...]

PR #383, opened 2019-11-20
More Data Frame Implementations + Code Refactoring

SinghGursimran:
Added more Data Frame Implementations along with some Code Refactoring

#223 

For Testing:

```scala
import io.archivesunleashed._

RecordLoader.loadArchives("./ARCHIVEIT-227-QUARTERLY-XUGECV-20091218231727-00039-crawling06.us.archive.org-8091.warc.gz",sc)
			.keepImages()
			.all()
			.select(ComputeImageSizeDF($"bytes").as("size"), ComputeMD5DF($"bytes").as("md5"), ComputeSHA1DF($"bytes").as("sha1"))
			.show(20,false)
```


src/main/scala/io/archivesunleashed/DataFrameLoader.scala
```diff
@@ -27,6 +27,12 @@ class DataFrameLoader(sc: SparkContext) {
       .pages()
   }
 
+  def pagesWithBytes(path: String): DataFrame = {
```
ruebot: Let's get a doc comment for this method. You can just crib from the methods around it for the pattern.

src/main/scala/io/archivesunleashed/package.scala
```diff
@@ -115,6 +115,24 @@ package object archivesunleashed {
       sqlContext.getOrCreate().createDataFrame(records, schema)
     }
 
+    /*Creates a column for Bytes as well in Dataframe. 
+      Call KeepImages OR KeepValidPages on RDD depending upon the requirement before calling this method */
+    def pagesWithBytes(): DataFrame = {
```
ruebot: @lintool you like naming things, you good with this one, or you got something better? :smiley:

src/main/scala/io/archivesunleashed/package.scala
```diff
@@ -30,7 +30,7 @@ import io.archivesunleashed.matchbox.ExtractDate.DateComponent.DateComponent
 import java.net.URI
 import java.net.URL
 import org.apache.spark.sql.{DataFrame, Row, SparkSession}
-import org.apache.spark.sql.types.{IntegerType, StringType, StructField, StructType}
+import org.apache.spark.sql.types.{IntegerType, StringType, BinaryType, StructField, StructType}
```
ruebot: Let's get this in alphabetical order so Scalastyle is happy. Pedantic, sorry.

lintool:
With DF, we shouldn't need `pagesWithBytes`, we should be okay with the widest possible table (as many fields as we need) in a single method. Spark is supposed to do the optimization behind the scenes, i.e., if you don't `select` it, Spark should be smart to project away it early on.

Worth testing just to make sure.

SinghGursimran:
@lintool Initially, I added this for images. I need to keep images for analysis. Instead of adding it specific to images, I created a general method considering "bytes" might be required in the future for non-image analysis as well. 
Previous "pages" method filters out only valid pages without keeping images. 
Plus, that "pages" method is being used in many use cases where we actually require only valid pages. So, I thought of not changing it. 
If you want, I would change this "pagesWithBytes" to be specific for images.

src/main/scala/io/archivesunleashed/package.scala
ruebot: @SinghGursimran as per slack conversation, let's call this one `all`.

I can take care of renaming `pages` to `html` tomorrow or later tonight under the cover of a separate PR.

PR #375, opened 2019-11-10
Spark 3.0.0 + Java 11 support.

ruebot:
**GitHub issue(s)**: #356

# What does this Pull Request do?

Mostly `pom.xml` updates. Lots of cleanup, and updates. Big stuff, update to Apache Spark 3.0.0, and Java 11.

# How should this be tested?

- TravisCI
- Exhaustive regression testing; basically hit everything in the documentation, and make sure it works.
- [Testing gist](https://gist.github.com/ruebot/b4d565c2a0cdc5e72678366f6d616388) (updating it as I go through)

# Additional Notes:

- **BIG KICKER**
  - I **cannot** get this to work with `--packages`. We hit an ugly dependency wall really quick. If we exclude, all the tests fail really bad. If I explicitly include it as a dependency, same thing. Tests fail really bad.
- I'm going to leave this as a draft, and we shouldn't merge until there is an official Spark 3.0.0 release, and we make a decision on `--packages`.
- We'll squash this all down, and make a nice commit message when the time comes.

Issue #356, opened 2019-08-27
Java 11 support


ruebot:
From the Apache Spark mailing list:

```
Hi, All.

Thanks to your many many contributions,
Apache Spark master branch starts to pass on JDK11 as of today.
(with `hadoop-3.2` profile: Apache Hadoop 3.2 and Hive 2.3.6)

    https://amplab.cs.berkeley.edu/jenkins/view/Spark%20QA%20Test%20(Dashboard)/job/spark-master-test-maven-hadoop-3.2-jdk-11/326/
    (JDK11 is used for building and testing.)

We already verified all UTs (including PySpark/SparkR) before.

Please feel free to use JDK11 in order to build/test/run `master` branch and
share your experience including any issues. It will help Apache Spark 3.0.0 release.

For the follow-ups, please follow https://issues.apache.org/jira/browse/SPARK-24417 .
The next step is `how to support JDK8/JDK11 together in a single artifact`.

Bests,
Dongjoon.
```

We'll align with Apache Spark here on Java 11 support. Once we have a Spark release with Java 11, I'll pivot to getting `aut` stable with Java 11.

ruebot:
Getting closer to Spark 3.0.0!

```
Hi all,

To enable wide-scale community testing of the upcoming Spark 3.0 release, the Apache Spark community has posted a preview release of Spark 3.0. This preview is not a stable release in terms of either API or functionality, but it is meant to give the community early access to try the code that will become Spark 3.0. If you would like to test the release, please download it, and send feedback using either the mailing lists or JIRA.

There are a lot of exciting new features added to Spark 3.0, including Dynamic Partition Pruning, Adaptive Query Execution, Accelerator-aware Scheduling, Data Source API with Catalog Supports, Vectorization in SparkR, support of Hadoop 3/JDK 11/Scala 2.12, and many more. For a full list of major features and changes in Spark 3.0.0-preview, please check the thread(http://apache-spark-developers-list.1001551.n3.nabble.com/Spark-3-0-preview-release-feature-list-and-major-changes-td28050.html).

We'd like to thank our contributors and users for their contributions and early feedback to this release. This release would not have been possible without you.

To download Spark 3.0.0-preview, head over to the download page: https://archive.apache.org/dist/spark/spark-3.0.0-preview

Thanks,

Xingbo
```

ruebot:
Successful build:
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[...]

Issue #329, opened 2019-07-24
Upgrade to Hadoop 3.x


jrwiebe:
AUT currently uses Hadoop 2.6.5. Though it is stable, at three years old it is beginning to show its age. I discovered this when testing S3 access (#319): hadoop-aws 2.6.5 is incapable of authenticating with temporary security credentials (probably an edge case) and with endpoints that require Signature Version 4  (many do). Upgrading to a current branch of Hadoop should be a matter of bringing other dependencies up to date, which might not be simple.

At present I would suggest going with version 3.1.2, the latest 3.1.x release. Its [docs say](https://hadoop.apache.org/docs/r3.1.2/): "This release is generally available (GA), meaning that it represents a point of API stability and quality that we consider production-ready." Or 3.0.0 -- all the [Cloudera CDH 6 releases](https://www.cloudera.com/documentation/enterprise/6/release-notes/topics/rg_cdh_6_packaging.html) use that version, which is an indication its stability and wide use.

I'm not sure of the implications of using a distribution of Spark built with an older Hadoop to run our code that depends on Hadoop 3 (Spark 2.4.3 uses Hadoop 2.6.5). I wonder how the version conflicts would be resolved if we included the Hadoop 3 dependencies in our fatjar (we currently exclude them), and run it on Spark with Hadoop 2.6.5? I imagine it should work if we include Hadoop in the fatjar if we instruct people to use the version of Spark built without Hadoop. I think it's unreasonable to expect people to build Spark themselves, though.

ruebot:
We'll have to sort out `FileUtil.copyMerge` since it was deprecated in 3+ versions of Hadoop. Luckily it is only used [here](https://github.com/archivesunleashed/aut/blob/19b49e11ed106b009ea601050423652718178855/src/main/scala/io/archivesunleashed/app/NERCombinedJson.scala#L63-L64).

Preliminary [StackOverFlow searching](https://stackoverflow.com/questions/42035735/how-to-do-copymerge-in-hadoop-3-0) says we can implement our own version in Scala. Is that something we'd want to do, or should we look for a better solution to combine all the part files.

...and maybe there is a way to pull off a [`hdfs cat`](https://stackoverflow.com/questions/5700068/merge-output-files-after-reduce-phase) :man_shrugging: 

Anyway, I'll keep digging and check out the Hadoop 3.1.2 API docs.

jrwiebe:
The Scala re-implementation looks good. I'd use it.

[...]

PR #385, opened 2019-11-26
 Spark 3.0.0 + Hadoop 3.2 + Java 11 support.

ruebot:
**GitHub issue(s)**:

- #329 
- #356 

This PR is #375 + Hadoop 3.2 support.

# How should this be tested?

Same as #375 + It should be tested with [spark-3.0.0-preview-bin-hadoop3.2](https://archive.apache.org/dist/spark/spark-3.0.0-preview/spark-3.0.0-preview-bin-hadoop3.2.tgz).

# Additional Notes:

Same as #375


src/main/scala/io/archivesunleashed/util/copyMerge.scala
```diff
@@ -0,0 +1,60 @@
...
+package io.archivesunleashed.util
+
+import scala.util.Try
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.{FileSystem, Path}
+import org.apache.hadoop.io.IOUtils
+import java.io.IOException
+
+/** Reimplementation of FileUtil.copyMerge . */
+object CopyMerge {
```
ruebot, line 31: @lintool we're adding this, since `copyMerge` was deprecated and removed from Hadoop 3. Do we need to actually implement it again, or should we just use `coalesce` like we do [here](https://github.com/archivesunleashed/aut/blob/master/src/main/scala/io/archivesunleashed/app/CommandLineApp.scala#L192-L224)?

lintool: I don't recall this being my code... but I think that's fine.

PR #389, opened 2019-12-16
Setup for Serializable APIs on DataFrames

SinghGursimran:
Setup for Serializable APIs on DataFrames

#223  

For Testing:
```scala

import io.archivesunleashed._

RecordLoader.loadArchives("./src/test/resources/arc/example.arc.gz",sc)
			.all()
			.keepValidPagesDF()
			.select($"crawl_date",$"url")
			.show(10,false)

```

Since it is the migration of RDD's functionality to DF, final answer should be the same as given by the following transformation on RDD.

```scala
import io.archivesunleashed._

RecordLoader.loadArchives("./src/test/resources/arc/example.arc.gz",sc)
			.webpages()
			.select($"crawl_date",$"url")
			.show(10,false)
```


SinghGursimran:
> @SinghGursimran got tests? 😃

oh! sorry forgot about that. I will add it now.

PR #397, opened 2020-01-02
WriteGraph DataFrame implementation

SinghGursimran:
WriteGraph DataFrame implementation

#223 

Reference to: https://github.com/archivesunleashed/aut-docs/pull/34
```scala

import io.archivesunleashed._

val target = udf((vs: Any) => {
								var res = ""
								if(vs != null){
									res = vs.toString.split(",")(1)}
								res
							})
val src = udf((vs: Any) => {
							var res = ""
							if(vs != null){
								val s = vs.toString.split(",")(0)
								if(s.length() != 0)
									res = s.drop(1)}
							res
						})	
val modify = udf((str: String) => str.replaceAll("^\\\\s*www\\\\.", ""))

var df = RecordLoader.loadArchives("./src/test/resources/arc/example.arc.gz",sc).webpages()
					 .select($"crawl_date",explode_outer(ExtractLinksDF($"url",$"content")).as("link"))


df = df.select($"crawl_date",modify(ExtractDomainDF(src($"link"))).as("Source"),modify(ExtractDomainDF(target($"link"))).as("Target"))
.filter($"Source" =!= "")
.filter($"Target" =!= "")
.groupBy("crawl_date","Source","Target")
.count()
.filter($"count" > 5)
.orderBy(desc("count"))

val Columns = Seq("crawl_date","Source","Target","count")
df = df.toDF(Columns:_*)

```

Step 1: Creating Nodes with Id from DataFrame

```scala
WriteGraph.nodesWithIdsDF(df).show(20)
```

Step 2: Creating Edges from Nodes and Dataframe
```scala
WriteGraph.edgeNodesDF(df).show(10)
```

Step 3: Writing Graph
```scala
WriteGraph.asGraphmlDF(df,"output.csv")
```

Remaining Tests.


SinghGursimran:
Yes, implementation is completed. I still have to find a way to design test cases for it.

Issue #408, opened 2020-01-17
Implement Python versions of Matchbox utilities


ruebot:
RDD | Scala DF | Python DF |
--- | --- | --- |
ComputeImageSize || |
ComputeMD5RDD  | ComputeMD5DF| in progress |
ComputeSHA1RDD | ComputeSHA1DF| in progress |
DetectLanguageRDD | DetectLanguageDF| in progress |
DetectMimeTypeTika |  DetectMimeTypeTikaDF||
ExtractBoilerPipeTextRDD | ExtractBoilerPipeTextDF||
ExtractDateRDD | ExtractDateDF||
ExtractDomainRDD | ExtractDomainDF| :heavy_check_mark: |
ExtractImageDetails ||
ExtractImageLinksRDD | ExtractImageLinksDF||
ExtractLinksRDD | ExtractLinksDF||
ExtractTextFromPDFs | -||
GetExtensionMimeRDD | GetExtensionMimeDF||
RemoveHTMLRDD | RemoveHTMLDF| :heavy_check_mark: |
RemoveHTTPHeaderRDD | RemoveHTTPHeaderDF| :heavy_check_mark: |
NERClassifier | -||
RemovePrefixWWW | RemovePrefixWWWDF| :heavy_check_mark: |

Stealing @SinghGursimran's very helpful tables here :smiley:

PR #412, opened 2020-01-18
Add some PySpark udfs

ruebot:
**GitHub issue(s)**: #408

- Add remove_http_header, remove_prefix_www
- Rename extract_domain_func to extract_domain
- Formatting updates
- Addresses #408

# How should this be tested?

https://gist.github.com/ruebot/e50892b0b2b4a6abad8ddc7933cf79b2

# Additional Notes:

* We need to sort out how we'll bundle something like everything that `requirements.txt` would help with in the zip file; something like the Uberjar. Right now, we really don't have anything, but I imagine we'd want to pull in external libraries like Beautiful Soup, or tld-extractor.

* I'll leave this as a draft, and push to it as I'm working on it. Others are welcome to push to it as well, since GitHub is now setup to provide credit to all those accounts who contributed to a PR when it is squashed down.

# Interested parties

@SinghGursimran if you're sick of Scala, let me know, and I can give you access.

@lintool this approach fine? Naming convention fine?

@ianmilligan1 let me know if the notebook testing makes sense. Figured that'd be easy to test this stuff.


SinghGursimran:
@ruebot, yes, sure, pls give the access.

ianmilligan1:
>@ianmilligan1 let me know if the notebook testing makes sense. Figured that'd be easy to test this stuff.

Makes sense to me! Thanks @ruebot.

src/main/python/aut/udfs.py
```diff
@@ -1,13 +1,75 @@
+import hashlib
+
+from bs4 import BeautifulSoup
 from pyspark.sql.functions import udf
 from pyspark.sql.types import StringType
+from textblob import TextBlob
+
+
+def compute_MD5(bytes):
```
ruebot: I'm getting a `TypeError: Unicode-objects must be encoded before hashing` on this and `compute_MD5`.

src/main/python/aut/udfs.py
```diff
@@ -1,13 +1,75 @@
...
+def compute_SHA1(bytes):
+    return hashlib.sha1(bytes).hexdigest()
+
 
+compute_SHA1 = udf(compute_SHA1, StringType())
 
-def extract_domain_func(url):
+
+def detect_language(input):
+    text = TextBlob(input)
```
ruebot: This doesn't look like it's going to scale: "[Language translation and detection powered by Google Translate](https://textblob.readthedocs.io/en/dev/#features)"

I ran it twice in testing on the 10 Geocities files I use locally, and I got hit with a: `urllib.error.HTTPError: HTTP Error 429: Too Many Requests.`

Which all begs a bigger question on porting a lot of these over to Python; do we need to? We provide an MD5, SHA1, and Language column now. Is there a use case for having them in Python? I can't think of a reason to run them on a column if we already provide them. @ianmilligan1 @lintool what do you think?

src/main/python/aut/udfs.py
SinghGursimran: Oh! It was working on my system...

ruebot: Interesting. What version of Python are you running? ...and you're on Windows too iirc?

SinghGursimran: I testes it on Linux, there I have python 2.8. I will test on windows as well. I have python 3.6 there.

ruebot: Ah, ok. I'm on 3.7.3 on Linux. I wonder if that's it. (Using the Anaconda distribution of Python)

SinghGursimran: Oh ok, I will update python to 3.7 and test with it.

src/main/python/aut/udfs.py
ianmilligan1: > I can't think of a reason to run them on a column if we already provide them. @ianmilligan1 @lintool what do you think?

Agreed with you - makes sense to not port in these cases (and especially in the particular case here).

src/main/python/aut/udfs.py
SinghGursimran: Hi @ruebot,
Did you try running this in the terminal. There's a timeout issue in jupyter notebook.

Something similar to what's mentioned here: https://www.idaima.org/topic/2447498/getting-error-caused-by-java-net-sockettimeoutexception-accept-timed-out/2

It is working in the terminal. I could not find any other library in python that would give the MD5 or SHA1 hashes. Currently, I am trying to make it work in jupyter.

PR #427, opened 2020-03-06
Restyle UDFs in the context of DataFrames

SinghGursimran:
Restyle UDFs in the context of DataFrames.

#425 

```scala
import org.apache.spark.sql.functions.lit
import io.archivesunleashed._
import io.archivesunleashed.df._

RecordLoader.loadArchives("./src/test/resources/arc/example.arc.gz",sc)
			.all()
			.select($"url",$"http_status_code")
			.filter(hasHttpStatus($"http_status_code", lit(Array("200","000"))))
			.show(10,false)

import org.apache.spark.sql.functions.lit
import io.archivesunleashed._
import io.archivesunleashed.df._

RecordLoader.loadArchives("./src/test/resources/arc/example.arc.gz",sc)
			.all()
			.select($"content")
			.filter(hasContent($"content", lit(Array("Content-Length: [0-9]{4}"))))
			.show(3,false)

import org.apache.spark.sql.functions.lit
import io.archivesunleashed._
import io.archivesunleashed.df._

RecordLoader.loadArchives("./src/test/resources/arc/example.arc.gz",sc)
			.all()
			.select($"url")
			.filter(hasUrlPatterns($"url", lit(Array(".*images.*"))))
			.show(10,false)

import org.apache.spark.sql.functions.lit
import io.archivesunleashed._
import io.archivesunleashed.df._

RecordLoader.loadArchives("./src/test/resources/arc/example.arc.gz",sc)
			.all()
			.select(DetectLanguageDF(RemoveHTMLDF($"content")).as("language"))
			.filter(hasLanguages(DetectLanguageDF(RemoveHTMLDF($"content")), lit(Array("de","ht"))))
			.show(10,false)

import org.apache.spark.sql.functions.lit
import io.archivesunleashed._
import io.archivesunleashed.df._

RecordLoader.loadArchives("./src/test/resources/arc/example.arc.gz",sc)
			.all()
			.select($"mime_type_web_server")
			.filter(hasMimeTypes($"mime_type_web_server", lit(Array("text/html"))))
			.show(10,false)

import org.apache.spark.sql.functions.lit
import io.archivesunleashed._
import io.archivesunleashed.df._

RecordLoader.loadArchives("./src/test/resources/arc/example.arc.gz",sc)
			.all()
			.select($"url")
			.filter(hasUrls($"url", lit(Array("http://www.archive.org/","http://www.archive.org/robots.txt"))))
			.show(10,false)

import org.apache.spark.sql.functions.lit
import io.archivesunleashed._
[...]

PR #429, opened 2020-03-17
Restyle keep/discard filter UDFs in the context of DataFrames

ruebot:
**GitHub issue(s)**: 
  - #425 
  - Supersedes #427 

Restyle keep/discard filter UDFs in the context of DataFrames

# How should this be tested?

- TravisCI
- https://gist.github.com/ruebot/0e4ca913cbce3cd02f228264d3638aa2

# Additional Notes:

- @SinghGursimran we still have a bit more to go here I think. I'll loop back around to this tomorrow morning, and double check. But, I _think_ it's just `hasImages`, and `hasMimeTypeTika`.
- I'll start a documentation PR for this tomorrow


src/main/scala/io/archivesunleashed/package.scala
```diff
@@ -89,230 +89,6 @@ package object archivesunleashed {
...
-
-    def hasColumn(df: DataFrame, path: String) = Try(df(path)).isSuccess
-
-    val spark = SparkSession.builder().master("local").getOrCreate()
-    // scalastyle:off
-    import spark.implicits._
-    // scalastyle:on
-
-    /** Removes all non-html-based data (images, executables, etc.) from html text. */
-    def keepValidPagesDF(): DataFrame = {
```
ruebot: We should probably keep some version of this. @lintool @ianmilligan1 in the spirit of #425, we should change this. Should it just be `hasValidPages`?

src/main/scala/io/archivesunleashed/package.scala
```diff
@@ -89,230 +89,6 @@ package object archivesunleashed {
...
-      *
-      * @param lang a set of ISO 639-2 codes
-      */
-    def discardLanguagesDF(lang: Set[String]): DataFrame = {
-      val filteredLanguage = udf((language: String) => !lang.contains(language))
-      df.filter(filteredLanguage(DetectLanguageDF(RemoveHTMLDF($"content"))))
-    }
-
-    /** Removes all data except images. */
-    def keepImagesDF(): DataFrame = {
```
ruebot: @ianmilligan1 @lintool `hasImages()` work for y'all?

src/main/scala/io/archivesunleashed/package.scala
ianmilligan1: That works for me!

src/main/scala/io/archivesunleashed/package.scala
ianmilligan1: Perfect!

src/main/scala/io/archivesunleashed/package.scala
ruebot: Well, maybe that's not the right path. `keepValidPagesDF()` is different that all of the others. Maybe it should stay as is, but be moved from `package.scala` to `df/package.scala`, and rename to `.keepValidPages()`. But, we might run into problems there because of the existing `keepValidPages()` in `package.scala`. Is it worth moving stuff of functions in `package.scala` to a new `rdd/package.scala`?

ruebot: ...or, just restore this small class with `keepValidPagesDF` and say it's not in scope for issue/PR?

src/main/scala/io/archivesunleashed/df/package.scala
```diff
@@ -59,6 +60,38 @@ package object df {
 
   val DetectMimeTypeTikaDF = udf((io.archivesunleashed.matchbox.DetectMimeTypeTika.apply(_: Array[Byte])))
 
+  val hasHTTPStatus = udf((statusCode: String, statusCodes: Seq[String]) => statusCodes.contains(statusCode))
+
+  val hasMIMETypes = udf((mimeType: String, mimeTypes: Seq[String]) => mimeTypes.contains(mimeType))
+
+  val hasMIMETypesTika = udf((mimeTypeTika: String, mimeTypes: Seq[String]) => mimeTypes.contains(mimeTypeTika))
```
SinghGursimran: Here, I think if we follow the convention of other UDF's, it should be "mimeTypesTika: Seq[String]".
I had by mistake used the wrong convention for MIMETypesTika while writing serializable APIs previously.

ruebot: Yeah, we've flipped between `MIME` and `Mime` in these. I'd say `Mime` looks better here, even though we changed `Http` to `HTTP`. @lintool what would you like here?

SinghGursimran: I meant, currently it is "mimeTypes: Seq[String]" but it should be "mimeTypesTika: Seq[String]" in the udf.

ruebot: Ah!

ruebot: Updated

PR #430, opened 2020-03-23
Tweak hasDate to handle Seq.

ruebot:
**GitHub issue(s)**:

* ex: #425
* ex: #429 

Follow on to #429, Tweak `hasDate` to handle Seq.

# How should this be tested?

- TravisCI

...and if you want, something like:

```scala
import io.archivesunleashed._

val dates = Array("2008", "200908", "20070502")

RecordLoader.loadArchives("/home/nruest/Projects/aut/sample-data/geocities", sc)
  .all()
  .select($"url", $"crawl_date")
  .filter(!hasDate($"crawl_date", lit(dates)))
  .show(10, false)  
```


src/main/scala/io/archivesunleashed/df/package.scala
```diff
@@ -66,7 +66,7 @@ package object df {
 
   val hasMIMETypesTika = udf((mimeType: String, mimeTypesTika: Seq[String]) => mimeTypesTika.contains(mimeType))
 
-  val hasDate = udf((date_ : String, date: String) => date_ == date)
+  val hasDate = udf((date_ : String, date: Seq[String]) => date_ == date)
```
SinghGursimran: Here, the condition will be => date.contains(date_)

ruebot: :man_facepalming: 

Good catch @SinghGursimran!

PR #453, opened 2020-04-21
Update PlainTextExtractor to output a single column; text.

### LINKED ISSUE #452: Update PlainTextExtractor to just extract text
Currently there is a fair bit of overlap between the [`PlainTextExtractor`](https://github.com/archivesunleashed/aut/blob/f1eb43bfdfe1fb08d4329c2140cc5934dfbe712c/src/main/scala/io/archivesunleashed/app/PlainTextExtractor.scala) and [`WebPagesExtractor`](https://github.com/archivesunleashed/aut/blob/f1eb43bfdfe1fb08d4329c2140cc5934dfbe712c/src/main/scala/io/archivesunleashed/app/WebPagesExtractor.scala). Really, the only different between them now is the name of the content/text column, and `WebPagesExtractor` has some additional columns.

I propose that `PlainTextExtractor` moves to something that is more in the spirit of its name. It should run `RemoveHTMLDF`, `RemoveHTTPHeaderDF`, a DataFrame version of `ExtractBoilerpipeTextRDD`, and output a single column (csv or parquet), or possibly a single text file.

ruebot:
**GitHub issue(s)**: #452 

Update PlainTextExtractor to output a single column; text.

- Resolves #452
- PlainTextExtractor runs RemoveHTML, and ExtractBoilerplate on
`content`
- Update test

# How should this be tested?

- TravisCI

```shell
bin/spark-submit --class io.archivesunleashed.app.CommandLineAppRunner /home/nruest/Projects/au/aut/target/aut-0.60.1-SNAPSHOT-fatjar.jar --extractor PlainTextExtractor --input /home/nruest/Projects/au/sample-data/geocities/* --output /home/nruest/Projects/au/sample-data/452-test/plaintext
```

### DIFF
--- a/src/main/scala/io/archivesunleashed/app/PlainTextExtractor.scala
+++ b/src/main/scala/io/archivesunleashed/app/PlainTextExtractor.scala
@@ -32,6 +32,6 @@ object PlainTextExtractor {
     // scalastyle:off
     import spark.implicits._
     // scalastyle:on
-    d.select(ExtractBoilerpipeTextDF(RemoveHTMLDF($"content")).as("content"))
+    d.select(ExtractBoilerpipeTextDF($"content").as("content"))
   }
 }
--- a/src/main/scala/io/archivesunleashed/app/PlainTextExtractor.scala
+++ b/src/main/scala/io/archivesunleashed/app/PlainTextExtractor.scala
@@ -17,8 +17,7 @@
 package io.archivesunleashed.app
 
 import io.archivesunleashed.ArchiveRecord
-import io.archivesunleashed.df.{ExtractBoilerpipeTextDF, RemoveHTMLDF,
-                                RemoveHTTPHeaderDF}
+import io.archivesunleashed.df.{ExtractBoilerpipeTextDF}
 import org.apache.spark.sql.{DataFrame, Dataset, Row, SparkSession}
 
 object PlainTextExtractor {
--- a/src/main/scala/io/archivesunleashed/app/PlainTextExtractor.scala
+++ b/src/main/scala/io/archivesunleashed/app/PlainTextExtractor.scala
@@ -17,7 +17,7 @@
 package io.archivesunleashed.app
 
 import io.archivesunleashed.ArchiveRecord
-import io.archivesunleashed.df.{ExtractDomainDF, RemoveHTMLDF,
+import io.archivesunleashed.df.{ExtractBoilerpipeTextDF, RemoveHTMLDF,
                                 RemoveHTTPHeaderDF}
 import org.apache.spark.sql.{DataFrame, Dataset, Row, SparkSession}
 
@@ -32,7 +32,6 @@ object PlainTextExtractor {
     // scalastyle:off
     import spark.implicits._
     // scalastyle:on
-    d.select($"crawl_date", ExtractDomainDF($"url").as("domain"),
-      $"url", RemoveHTMLDF(RemoveHTTPHeaderDF($"content")).as("text"))
+    d.select(ExtractBoilerpipeTextDF(RemoveHTMLDF($"content")).as("content"))
   }
 }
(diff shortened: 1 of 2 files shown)


ianmilligan1:
Oh, I'm using data from the CPP collection – and doesn't appear to have any data in the whole collection, whereas yes, there's like two records that come out in the text extractor. There's a lot of junk records you'd expect to see removed, but there are some legit URLs that aren't coming through. i.e. I'm getting 46KB of text whereas running the old `WebPages` is 25MB (albeit with more columns).

Let me poke at this a bit - I am pretty sure I've run BoilerPipe on these WARCs before and the results have been bigger than this.

src/main/scala/io/archivesunleashed/app/PlainTextExtractor.scala
```diff
@@ -32,7 +32,6 @@ object PlainTextExtractor {
     // scalastyle:off
     import spark.implicits._
     // scalastyle:on
-    d.select($"crawl_date", ExtractDomainDF($"url").as("domain"),
-      $"url", RemoveHTMLDF(RemoveHTTPHeaderDF($"content")).as("text"))
+    d.select(ExtractBoilerpipeTextDF(RemoveHTMLDF($"content")).as("content"))
```
ruebot, line 35: https://github.com/archivesunleashed/aut/blob/f1eb43bfdfe1fb08d4329c2140cc5934dfbe712c/src/main/scala/io/archivesunleashed/matchbox/ExtractBoilerpipeTextRDD.scala#L31-L33

We really don't have any documentation on using that from what I can tell.

Maybe we shouldn't be calling `RemoveHTMLDF` before calling `ExtractBoilerpipeTextDF` here?

@ianmilligan1 @lintool

ianmilligan1:
Tested with the DF script from [here](https://github.com/archivesunleashed/aut-docs/blob/master/current/text-analysis.md#Extract-Plain-Text-Minus-Boilerplate):

```scala
import io.archivesunleashed._

RecordLoader.loadArchives("/Users/ianmilligan1/dropbox/git/aut-resources/Sample-Data/*.gz", sc)
  .webpages()
  .select($"crawl_date", ExtractDomainDF($"url"), $"url", ExtractBoilerpipeTextDF(RemoveHTTPHeaderDF($"content")))
  .write.csv("plain-text-no-boilerplate-df-testing-453/")
```

And yes results are more robust (i.e. from scrolling through the CSV at a glance obvious boilerplate has been removed but content is still there in many cases; exponentially more than when using the `PlainTextExtractor` app).

src/main/scala/io/archivesunleashed/app/PlainTextExtractor.scala
ianmilligan1: Yeah, I think you're right on that @ruebot. From comparing our docs, when we do regular text extract in DF, we use a command like:

```scala
  .select($"crawl_date", ExtractDomainDF($"url").as("domain"), $"url", RemoveHTMLDF(RemoveHTTPHeaderDF($"content")).as("content"))
```

so yes the headers are removed and then the HTML is removed. Whereas in the boilerpipe we use:

```scala
  .select($"crawl_date", ExtractDomainDF($"url"), $"url", ExtractBoilerpipeTextDF(RemoveHTTPHeaderDF($"content")))
```

So we omit the HTML step and run Boilerpipe on the HTTP-headerless content.

ruebot: Cool. It should just be `ExtractBoilerpipeTextDF` then, since that calls `ExtractBoilerpipeTextRDD` which runs `RemoveHTTPHeaderRDD` before running `removeBoilerplate`.

I'll update this, the test, and push it up shortly.

ianmilligan1: Perfect! And I've got the output now from the shell comparator, so I can quickly see how it more or less lines up.

Issue #468, opened 2020-05-27
PEP8 Naming  - UDFs, App method names, DataFrame names, and filters.


ruebot:
Let's evaluate how we've named everything; UDFs, App method names, DataFrame names, and filters.

Are we all good with how we've named everything?

Do we need to change anything before a 1.0.0 release?

See also #467

lintool:
My thoughts: In Python, I would default to PEP8 unless there's a convincing reason otherwise: https://www.python.org/dev/peps/pep-0008/

Apps are most like Classes, so I would go CamelCase. Note, it says:

> Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.

UDFs are like methods, so should follow standard snake_case.

ruebot:
@lintool anything on the Scala side? If not, I'll get some PRs ready for aut, documentation, and notebooks.

ruebot:
- `ExtractPopularImages`
- `WriteGexf`
- `WriteGraphml`

ruebot:
Resolved with https://github.com/archivesunleashed/aut/commit/8037745308566003a6b1939e3dafa154bc5b5db5

lintool:
Feel free to vote down if you think this ship has already sailed... Currently, we have `removeHTML` and `remove_html` for DF UDFs in Scala and Python, respective. This is good. 

However, for RDDs, we still have `RemoveHTMLRDD`. I suppose we can rename those to just `RemoveHTML`. The casing differences make it clear that small camel is DF and regular camel is RDD?

lintool:
And also: https://github.com/archivesunleashed/aut-docs/blob/master/current/dataframe-schemas.md

Thoughts about Python/Scala naming conventions? Right now we use small camel case. In Python DF, this goes against the best practice of downcased snake case.

There are three relevant cases:
+ `.presentationProgramFiles()`
+ `.wordProcessorFiles()`
+ `.textFiles()`

If we change to:
+ `.presentations()`
+ `.documents()`
+ `.textfiles()`

We would completely resolve any of these issues. `.documents()` might be slightly confusing; `.textfiles()` seems fine to me as all lower cased since we have `.imagegraph()`.

---

And with this, I think I've said my peace on naming.

ruebot:
Good call on the Scala side. It'll be nice to remove those RDD suffixes. I'll get the Scala side sorted out here in a bit.

[...]

PR #479, opened 2020-06-01
Remove RDD suffixes on file, class, and object names.

ruebot:
**GitHub issue(s)**: #468

Remove RDD suffixes on file, class, and object names.

- Remove all the RDD suffixes added previously
- Rename image_graph to imagegraph (Python)
- Rename GetExtensionMime to GetExtensionMIME (Scala)
- Remove textFiles (Scala)
- Remove text_files (Python)
- Remove TextFilesInformationExtractor
- Rename files all affected files as needed
- Update tests as needed

# How should this be tested?

- TravisCI _should_ do it
- But, a good sniff test would be good as well.

# Additional Notes:

- Since there is so much documentation changes, I'll get the documentation updated on the docusaurus updates I'm working on. I'll try and push up a branch there with the changes today.
- Notebooks will need to be updated. I'll try and get that updated later today as well


src/main/scala/io/archivesunleashed/matchbox/GetExtensionMime.scala
```diff
@@ -18,9 +18,7 @@ package io.archivesunleashed.matchbox
 import org.apache.commons.io.FilenameUtils
 
 /** Get file extension using MIME type, then URL extension. */
-// scalastyle:off object.name
-object GetExtensionMimeRDD {
-// scalastyle:on object.name
+object GetExtensionMime {
```
lintool, line 21: Ah, caught this one - shouldn't this be `GetExtensionMIME`?

PR #520, opened 2021-10-04
Update ExtractDomain to extract apex domains.

### LINKED ISSUE #519: ExtractDomains returns non-Apex Domains
Not 100% sure this is the intention of `ExtractDomain` since it was implemented before my time, but I'm inclined to believe that it should return the domain apex, instead of sub-domains.

So, `aut.doc.archivesunleashed.org` and `news.archivesunleashed.org` should all be returned as `archivesunleashed.org`.

If non-apex domains were the original intention, let me know @lintool @ianmilligan1.

ruebot:
**GitHub issue(s)**: #519

- Resolves #519
- Add scala-uri as a dependency
- Replace getHost method of extracting domains with apexDomain from scala-uri
- Update tests as needed
- Removed unused source parameter from ExtractDomain

# How should this be tested?

* GitHub Actions should take care of it
* I'm going to test this at scale on tuna or orca
* We should test this with ARCH too https://github.com/helgeho/ars-cloud/issues/123

# Additional Notes:

1. I removed the `source` parameter since I honestly had no idea what it was for, or what the use case was. Couldn't really see anything in the commit history. @ianmilligan1 @lintool do y'all recall what it was for?

2. Once this is merged, I'll cut a new release since we'll need it for ARCH.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -551,6 +551,11 @@
       <artifactId>hadoop-aws</artifactId>
       <version>${hadoop.version}</version>
     </dependency>
+    <dependency>
+      <groupId>io.lemonlabs</groupId>
+      <artifactId>scala-uri_${scala.binary.version}</artifactId>
+      <version>3.5.0</version>
+    </dependency>
   </dependencies>
 
   <developers>
--- a/src/main/scala/io/archivesunleashed/matchbox/ExtractDomain.scala
+++ b/src/main/scala/io/archivesunleashed/matchbox/ExtractDomain.scala
@@ -15,6 +15,7 @@
  */
 package io.archivesunleashed.matchbox
 
+import io.lemonlabs.uri.Url
 import java.net.URL
 
 /** Extracts the host domain name from a full url string. */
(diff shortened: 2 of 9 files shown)


ianmilligan1:
>I removed the source parameter since I honestly had no idea what it was for, or what the use case was. Couldn't really see anything in the commit history. @ianmilligan1 @lintool do y'all recall what it was for?

No clue. Went back in the blame and looks like it's been there forever (maybe from when Alice was working on this five or six years ago). If I had to guess it was some janky error handling. I think yanking it makes sense.

ianmilligan1:
> Well, this blows up in Spark 3.0.3 and 3.

😞 

(that sucks! and odd behaviour... )

lintool:
> 1. I removed the `source` parameter since I honestly had no idea what it was for, or what the use case was. Couldn't really see anything in the commit history. @ianmilligan1 @lintool do y'all recall what it was for?

My memory is faint on this, but - per docs, "source an optional default url for urls with no valid domain host".

I think we've run into cases before where the full URL that we're trying to extract from was mangled. Usually, we're doing extraction on an URL on some page - so we set "source" to that page. Rationale is that if the URL is mangled, we can back up to the page that URL was found on.

That was the original rationale. As to whether it's a good one, that's a different matter...

PR #539, opened 2022-05-27
Make webpages() consistent across aut and ARCH.

### LINKED ISSUE #538: Remove http headers, and html on webpages()
In ARCH we remove headers and html on `.webpages()`. We should be consistent with none.

If folks need the content with headers and html, they can grab it from `.all()`.

ruebot:
**GitHub issue(s)**:

* #538

Make webpages() consistent across aut and ARCH.

- Filter HTTP headers, and HTML from content on webpages so that it is consistent with the app implementation, and the ARCH implementation
- Update PlainTextExtractor to use .all() since HTML is removed from content
- Add domain to all()
- Update csv exports on app so that they are rfc4180 compliant
- Apply GitHub workflows to main branch
- Consistent formating on DataFrameLoader.scala
- Update tests as needed
- Resolves #538

# How should this be tested?

* GitHub actions build
* Testing using documentation updates: https://github.com/archivesunleashed/aut-docs/pull/117

### DIFF
--- a/src/main/scala/io/archivesunleashed/app/CommandLineApp.scala
+++ b/src/main/scala/io/archivesunleashed/app/CommandLineApp.scala
@@ -20,7 +20,7 @@ import java.nio.file.{Files, Paths}
 
 import io.archivesunleashed.{ArchiveRecord, RecordLoader}
 import org.apache.log4j.Logger
-import org.apache.spark.sql.{Dataset, Row}
+import org.apache.spark.sql.{DataFrame, Dataset, Row}
 import org.apache.spark.{SparkConf, SparkContext}
 import org.rogach.scallop.exceptions.ScallopException
 import org.rogach.scallop.ScallopConf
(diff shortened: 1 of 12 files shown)


src/main/scala/io/archivesunleashed/app/CommandLineApp.scala
```diff
@@ -327,12 +327,19 @@ class CommandLineApp(conf: CmdAppConf) {
...
-        .csv(saveTarget)
+        .option("escape", "\"")
+        .option("encoding", "utf-8")
+        .save(saveTarget)
     } else {
       d.write
         .option("timestampFormat", "yyyy/MM/dd HH:mm:ss ZZ")
-        .csv(saveTarget)
+        .format("csv")
+        .option("header", "true")
```
ruebot, line 339: Oh, I need to remove this header option here.

ianmilligan1:
>Does that make sense?

Makes a lot of sense – I like `raw_content`. I can test the PR when it's updated and ready to merge!

ianmilligan1 (approved):
Tested and all works! (Except for the discovery of Spark 3.0.0 having a bug, as discussed on Slack!)

